PR #29, opened 2018-06-28
Formats

viktor76525:
new stuff

### DIFF
--- a/qcore/formats.py
+++ b/qcore/formats.py
@@ -1,0 +1,27 @@
+"""
+Functions and classes to load data that doesn't belong elsewhere.
+"""
+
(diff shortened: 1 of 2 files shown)


qcore/formats.py
```diff
@@ -0,0 +1,29 @@
...
+
+import numpy as np
+
+def load_im_file(csv_file, all_psa = False):
+
+    # process column names
+    use_cols = []
+    col_names = []
+    with open(csv_file, 'r') as f:
+        raw_cols = f.readline().strip().split(',')
```
sungeunbae: what do you think of using this instead?
f.readline().replace(" ","").split(',')

PR #39, opened 2018-12-12
Yaml

ttgenius:
Added class and functions in utils.py for the new cybershake install structure with yaml files in slurm_gm_workflow


qcore/utils.py
```diff
@@ -6,8 +6,149 @@
 from shutil import rmtree
 import os
 import imp
+import yaml
+from collections import OrderedDict
+from collections import Mapping
 
-def setup_dir(directory, empty = False):
+
+class DotDictify(dict):
```
viktor76525: you cannot use inbuilt functions as variables (dict)
https://docs.python.org/3.3/library/functions.html

viktor76525: is there any point to make a whole bunch of complexity just to do d.k instead of d['k']?? There needs to be a legitimate reason to do this

jasonmotha: Fair point. And that was discussed a while ago. For backwards / forwards compatibility reasons it was decided to support both in the first instance.

ttgenius: Not sure what it means. Could you please give an example

viktor76525: ok, here it doesn't do what I think it does. you are extending the dictionary class.
my question now is why is this code so complicated when a quick search comes up with this:
https://www.blog.pythonlibrary.org/2014/02/14/python-101-how-to-change-a-dict-into-a-class/

ttgenius: This class Dict2Obj does not work for nested dict like a={'b':2,'c':{'d':3,'z':4}}.
if you try aa=Dict2Obj(a); aa.c.d will throw an AttributeError

qcore/utils.py
```diff
@@ -6,8 +6,149 @@
...
+def ordered_load(stream, Loader=yaml.Loader, object_pairs_hook=OrderedDict):
+    """
+    :param stream: yaml file path
+    :param Loader: yaml loader
+    :param object_pairs_hook: always=OrderedDict to load file in order;
+
+    :return: OrderedDict
+    """
+
+    class OrderedLoader(Loader):
```
viktor76525: here you are extending the Loader class yet you haven't actually extended the Loader class.

qcore/utils.py
```diff
@@ -6,8 +6,149 @@
...
+    :return: OrderedDict/dict
+    """
+    with open(yaml_file, 'r') as stream:
+        try:
+            if obj_type is None:
+                return yaml.load(stream)
+            else:
+                return ordered_load(stream, Loader=yaml.SafeLoader, object_pairs_hook=obj_type)
+        except yaml.YAMLError as exc:
+            print(exc)
```
viktor76525: what does this except do? i'm not sure but does it show a part of the message (making debugging harder) instead of the whole trace message (useful for debugging) and therefore pointless.

qcore/utils.py
viktor76525: why do we have a complete mess of code that uses nested dicts?? can we rather make nice code that reduces complexity rather than adding sticky tape everywhere and making everything complex. This whole file is full of trivial yet messy wrapper functions that really shouldn't be necessary and I honestly feel sick trying to understand this spaghetti. 
Ask someone else to review, I refuse to deal with this.

qcore/utils.py
```diff
@@ -6,8 +6,146 @@
...
+
+    __setattr__, __getattr__ = __setitem__, __getitem__
+
+
+def ordered_load(stream, Loader=yaml.Loader, object_pairs_hook=OrderedDict):
+    """
+    :param stream: yaml file path
+    :param Loader: yaml loader
+    :param object_pairs_hook: always=OrderedDict to load file in order;
+
```
claudio525: why have object_pairs_hook as a parameter if its always the type OrderedDict?

ttgenius:
good point. Has made some changes to clarify.

PR #42, opened 2019-01-13
Qsw 683 - estimate wall clock

claudio525:
Python3 compatibility changes.

### DIFF
--- a/qcore/geo.py
+++ b/qcore/geo.py
@@ -128,11 +128,11 @@ def gp2ll_multi(coords, mlat, mlon, rot, nx, ny, hh):
                     'ylen=%s' % (ylen)],
                    stdin=PIPE, stdout=PIPE)
     stdout = p_conv.communicate(
-            '\n'.join(['%s %s' % tuple(c) for c in coords]).encode())[0]
+            '\n'.join(['%s %s' % tuple(c) for c in coords]).encode())[0].decode()
 
     # lon, lat
     return [list(map(float, line.split()))
-            for line in stdout.rstrip().split(b'\n')]
+            for line in stdout.rstrip().split('\n')]
 
 
 def gp2ll(x, y, mlat, mlon, rot, nx, ny, hh):
--- a/qcore/geo.py
+++ b/qcore/geo.py
@@ -2,24 +2,27 @@
 Various tools which may be needed in various processes.
 """
 
-from math import sin, asin, cos, atan, atan2, degrees, radians, sqrt, pi
 import os
+import numpy as np
 from subprocess import Popen, PIPE
 
-import numpy as np
+from math import sin, asin, cos, atan, atan2, degrees, radians, sqrt, pi
 
 from qcore.config import qconfig
+
 ll2xy_bin = os.path.join(qconfig['tools_dir'], 'll2xy')
 xy2ll_bin = os.path.join(qconfig['tools_dir'], 'xy2ll')
 
 R_EARTH = 6378.139
 
+
 class InputError(Exception):
     pass
 
-def ll2gp_multi(coords, mlon, mlat, rot, nx, ny, hh, \
-        dx = 1, dy = 1, decimated = False, verbose = False, \
-        keep_outside = False):
+
+def ll2gp_multi(coords, mlon, mlat, rot, nx, ny, hh,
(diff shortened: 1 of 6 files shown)


qcore/geo.py
```diff
@@ -2,24 +2,27 @@
 Various tools which may be needed in various processes.
 """
 
-from math import sin, asin, cos, atan, atan2, degrees, radians, sqrt, pi
 import os
+import numpy as np
```
viktor76525: import builtin1 (os, sys, math, subprocess)
import builtin2

import external1 (numpy, h5py etc..)
import external2

import local1 (qcore or other local files)
import local2

i try to go with the above format, alphabetical within blocks, not sure why they have been rearranged

qcore/geo.py
```diff
@@ -2,24 +2,27 @@
...
 class InputError(Exception):
     pass
 
-def ll2gp_multi(coords, mlon, mlat, rot, nx, ny, hh, \
-        dx = 1, dy = 1, decimated = False, verbose = False, \
-        keep_outside = False):
+
+def ll2gp_multi(coords, mlon, mlat, rot, nx, ny, hh,
+                dx=1, dy=1, decimated=False, verbose=False,
+                keep_outside=False):
```
viktor76525: just run "black code.py" (https://github.com/ambv/black) and it will format without bothering about it

qcore/geo.py
```diff
@@ -38,16 +41,18 @@ def ll2gp_multi(coords, mlon, mlat, rot, nx, ny, hh, \
...
-    p_conv = Popen(cmd,
-            stdin = PIPE, stdout = PIPE)
-    stdout = p_conv.communicate( \
-            '\n'.join(['%s %s' % tuple(c) for c in coords]))[0]
-    xy = [map(float, line.split()) for line in stdout.rstrip().split('\n')]
+
+    # Has to be a byte string
+    p_conv = Popen(cmd, stdin=PIPE, stdout=PIPE)
+    stdout = p_conv.communicate(
+        '\n'.join(['%s %s' % tuple(c) for c in coords]).encode())[0]
```
viktor76525: this will produce a byte array. need to run .decode() after too.

qcore/geo.py
```diff
@@ -38,16 +41,18 @@ def ll2gp_multi(coords, mlon, mlat, rot, nx, ny, hh, \
...
+    xy = [list(map(float, line.split()))
+          for line in stdout.rstrip().split(b'\n')]
```
viktor76525: i guess considering the final output is actually a float, converted then it's ok to not run decode as long as this part works fine.

qcore/srf.py
```diff
@@ -9,21 +9,21 @@
...
-from math import ceil, cos, floor, radians, sqrt
 import os
-from subprocess import Popen, PIPE
-
 import numpy as np
 
-from qcore import geo
-from qcore.config import qconfig
+from math import ceil, cos, floor, radians, sqrt, sin, degrees, atan
+from subprocess import Popen, PIPE
```
viktor76525: again, the built in python modules (os, math, subprocess) have been put in separate blocks when they belong together.

qcore/geo.py
sungeunbae: Sounds good, but should we enforce this rule? I think we need to have a set of internal rules that everyone can agree and follow

qcore/geo.py
```diff
@@ -115,22 +122,26 @@ def gp2ll_multi(coords, mlat, mlon, rot, nx, ny, hh):
...
-            stdin = PIPE, stdout = PIPE)
-    stdout = p_conv.communicate( \
-            '\n'.join(['%s %s' % tuple(c) for c in coords]))[0]
+    p_conv = Popen([xy2ll_bin, 'mlat=%s' % (mlat), 'mlon=%s' % (mlon),
+                    'geoproj=1', 'center_origin=1', 'h=%s' % (hh),
+                    'xazim=%s' % (xazim), 'xlen=%s' % (xlen),
+                    'ylen=%s' % (ylen)],
+                   stdin=PIPE, stdout=PIPE)
+    stdout = p_conv.communicate(
+            '\n'.join(['%s %s' % tuple(c) for c in coords]).encode())[0]
```
viktor76525: missed one (to keep consistent)

PR #54, opened 2019-02-01
Tools dir

ttgenius:
1. run with tools_dir branch of slurm_gm_worklfow
2. Added function to dynamically retrieve tools_dir in config.py
3. Original funcs in config.py is preserved to prevent breaks in qcore lib, but will progressively remove

### DIFF
--- a/qcore/config.py
+++ b/qcore/config.py
@@ -27,23 +27,23 @@ def get_host_config():
     :return: actual hostname, config josn file path
     """
     host_name = platform.node()
-
-    if host_name.startswith("ni") and len(host_name) == 8:  # maui
-        host_name = "maui"
+    if (host_name.startswith("ni") and len(host_name) == 8) or host_name.startswith('maui'):  # maui
+        actual_host_name = "maui"
         base_name = os.path.join('machine_config', 'config_maui.json')
 
     elif host_name.startswith("mahuika") and len(host_name) == 6:  # mahuika
-        host_name = "mahuika"
+        actual_host_name = "mahuika"
         base_name = os.path.join('machine_config', 'config_mahuika.json')
 
     else:  # default
+        actual_host_name = "default"
         base_name = "config.json"
 
     config_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), base_name)
-    return host_name, config_path
+    return actual_host_name, config_path
 
 
-def get_tools_dir(bin_name='emod3d', version='3.0.4-gcc'):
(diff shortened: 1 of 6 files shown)


qcore/config.py
```diff
@@ -19,3 +19,46 @@
...
     qconfig = json.load(f)
+
+
+def get_host_config():
+    """
+    Determine the actual hostname eg.ni0002--> maui; maui01-->maui
+    :return: actual hostname, config josn file path
+    """
+    host_name = platform.node()
+    if (host_name.startswith("ni") and len(host_name) == 8) or host_name.startswith('maui'):  # maui
```
jasonmotha: Len(host_name) isn't as portable as we'd like the qcore library to be.

qcore/config.py
```diff
@@ -19,3 +19,46 @@
...
+
+    else:  # default
+        actual_host_name = "default"
+        base_name = "config.json"
+
+    config_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), base_name)
+    return actual_host_name, config_path
+
+
+def get_tools_dir(bin_name, version='3.0.4-gcc'):
```
jasonmotha: This function doesn't seem to do what is described.

qcore/config.py
```diff
@@ -19,3 +21,46 @@
...
+    """
+    Determine the actual hostname eg.ni0002--> maui; maui01-->maui
+    :return: actual hostname, config json file path
+    """
+    host_name = platform.node()
+    if (host_name.startswith("ni") and len(host_name) == 8) or host_name.startswith('maui'):  # maui
+        actual_host_name = "maui"
+        base_name = os.path.join('machine_config', 'config_maui.json')
+
+    elif host_name.startswith("mahuika") and len(host_name) == 6:  # mahuika
```
jasonmotha: This case will never be true?

viktor76525:
@ttgenius please note the changes of config.py and update your usage accordingly
 - by design pattern, importer of config.py should not know where the json file is
 - there are different config files for different machines so that we don't have to make functions which do this, tools_dir points to tools dir and doesn't need to be tools + hostname + etc...
 - host variable is available so we do not need a duplicated function anymore if you would like access to this, instructions are at the top of config.py
 - tools_dir and opt_dir are no longer relevant because each host has a copy to the correct tools dir
 - if you want to find a location to a binary do: "{}/{}-{}".format(qconfig, "tool_name", "version") or similar (external to config.py)

please remember to git pull, thank you!

ttgenius:
not relevant any more

PR #58, opened 2019-02-07
Get bin ver

jpa198:


qcore/binary_version.py
```diff
@@ -0,0 +1,38 @@
+"""
+A class to retrieve versioned files.
+Defaults are available from the config files
+"""
+
+from config import qconfig
```
ttgenius: from qcore.config import qconfig otherwise might break?

jpa198: Should include this in the consistent coding guide?

PR #60, opened 2019-02-07
Change qcore to use the tool retrieval file.

jpa198:
Removed unused imports

### DIFF
--- a/qcore/gen_coords.py
+++ b/qcore/gen_coords.py
@@ -5,9 +5,9 @@ import os
 from subprocess import check_call
 import sys
 
-from qcore.config import qconfig
+from qcore.binary_version import get_unversioned_bin
 
-GEN_COORD_BIN = os.path.join(qconfig['tools_dir'], 'gen_model_cords')
+GEN_COORD_BIN = get_unversioned_bin('gen_model_cords')
 
 def gen_coords(vm_dir = '.', debug = False, geoproj = '1', do_coords = '1', \
         centre_origin = '1'):
--- a/qcore/geo.py
+++ b/qcore/geo.py
@@ -2,16 +2,15 @@
 Various tools which may be needed in various processes.
 """
 
-import os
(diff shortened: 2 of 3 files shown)


qcore/gen_coords.py
```diff
@@ -5,9 +5,9 @@
 from subprocess import check_call
 import sys
 
-from qcore.config import qconfig
+from qcore.binary_version import get_unversioned_bin
 
-GEN_COORD_BIN = os.path.join(qconfig['tools_dir'], 'gen_model_cords')
+GEN_COORD_BIN = get_unversioned_bin('gen_model_cords')
```
claudio525: Not sure if this should be a constant? From what I can see it is only used in one place (unless it used in other scripts), so might as well just change that to the function call?

qcore/geo.py
```diff
@@ -2,16 +2,15 @@
...
 
 import numpy as np
 
-from qcore.config import qconfig
+from qcore.binary_version import get_unversioned_bin
 
-ll2xy_bin = os.path.join(qconfig['tools_dir'], 'll2xy')
-xy2ll_bin = os.path.join(qconfig['tools_dir'], 'xy2ll')
+ll2xy_bin = get_unversioned_bin('ll2xy')
+xy2ll_bin = get_unversioned_bin('xy2ll')
```
claudio525: same as above

qcore/srf.py
```diff
@@ -9,16 +9,15 @@
...
 from subprocess import Popen, PIPE
 
 import numpy as np
 
-from qcore.config import qconfig
+from qcore.binary_version import get_unversioned_bin
 
 # binary paths
-srf2xyz = os.path.join(qconfig['tools_dir'], 'srf2xyz')
+srf2xyz = get_unversioned_bin('srf2xyz')
```
claudio525: same as above

PR #63, opened 2019-02-11
Sim struct yaml

ttgenius:
Modified config.py, binary_verison.py to work with different binary versions (slurm_gm_workflow tools_dir branch)
Modified shared.py to as type = file no longer supported by python3

### DIFF
--- a/qcore/config.py
+++ b/qcore/config.py
@@ -10,13 +10,13 @@ def determine_machine_config():
     :return: machine name, config file
     """
     hostname = platform.node()
-    if (hostname.startswith("ni") and len(host_name) == 8) or host_name.startswith(
+    if (hostname.startswith("ni") and len(hostname) == 8) or hostname.startswith(
         "maui"
     ):
         machine = "maui"
         basename = os.path.join("machine_config", "config_maui.json")
 
-    elif (host_name.startswith("wb") and len(host_name) == 6) or host_name.startswith(
+    elif (hostname.startswith("wb") and len(hostname) == 6) or hostname.startswith(
         "mahuika"
     ):
         machine = "mahuika"
--- a/qcore/config.py
+++ b/qcore/config.py
@@ -9,15 +9,15 @@ def get_host_config():
     :return: actual hostname, config josn file path
     """
     host_name = platform.node()
-    if (host_name.startswith("ni") and len(host_name) == 8) or host_name.startswith('maui'):  # maui
+    if (host_name.startswith("ni") and len(host_name) == 8) or host_name.startswith('maui'):
         actual_host_name = "maui"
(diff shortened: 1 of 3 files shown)


qcore/config.py
```diff
@@ -1,22 +1,34 @@
...
-    host = "mahuika"
-    basename = os.path.join('machine_config', 'config_mahuika.json')
-else:
-    # default
-    host = "unknown"
-    basename = 'config.json'
 
+def get_host_config():
+    """
+    Determine the actual hostname eg.ni0002--> maui; maui01-->maui
```
viktor76525: the actual hostname is the result of platform.node could say machine name or something

qcore/config.py
```diff
@@ -1,22 +1,34 @@
...
+    :return: actual hostname, config josn file path
+    """
+    host_name = platform.node()
+    if (host_name.startswith("ni") and len(host_name) == 8) or host_name.startswith('maui'):  # maui
```
viktor76525: comments at the end of lines are messy. pointless anyway because on the next line it says "maui"

qcore/simulation_structure.py
```diff
@@ -82,3 +82,33 @@ def get_hf_acc_dir(sim_root):
 def get_hf_bin_path(sim_root):
     return os.path.join(get_hf_acc_dir(sim_root), 'HF.bin')
 
+
+# yaml
+def get_fault_yaml_path(sim_root, fault_name=''):
+    """
+    manual: Albury_VM_home_melodypzhu_Albury_new_bench_Data_VMs_Albury-h0p4_EMODv3p0p4_190105/fault_params.yaml
```
jasonmotha: What does manual / auto mean in this case?

Can we use simplified examples for the description? I feel like this isn't actually clarifying what the script is doing.

claudio525: I actually think in general there should not be an example in the docstring at all and it should just be a description of what the function does. 

E.g. 
```
"""Gets the fault params yaml for the specified simulation. 

Note: For the manual workflow set fault_name to None as the fault params are stored directly in the simulation directory.
"""
```

Also I don't think the parameters need to be specified in this case as they are pretty self explanatory. If you want to specify them, then thats perfectly fine, but give a description instead of a example in my opinion.

E.g.
```
sim_root: Path to the `Runs` directory for the automated workflow, or the path the simulation directory for the manual workflow
fault_name: Name of the fault for automated workflow, None for manual workflow
```

Also on that note, I think the default value for fault_name should be None as in my opinion that is much clearer.
And then just add an extra line
```
fault_name = '' if fault_name is None else fault_name
```

ttgenius: Very good example, thanks. I have updated the comments

PR #87, opened 2019-03-14
Added common test set up and tear down functions to qcore

jpa198:
qcore is often a dependency of repositories that require unit tests, so this helps prevent code duplication.

### DIFF
--- a/qcore/testing.py
+++ b/qcore/testing.py
@@ -4,8 +4,24 @@ from shutil import rmtree
 
 from qcore import shared
 
+"""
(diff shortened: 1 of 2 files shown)


qcore/testing.py
```diff
@@ -0,0 +1,47 @@
+import sys
+import os
+from shutil import rmtree
+
+from qcore import shared
+
+
+def test_set_up(realizations):
```
claudio525: Maybe some docstring for a brief overview of what this function does? And where its used?

PR #102, opened 2019-03-29
Moved create_directory() from seisfinder site_utils.py to qcore utils.py

sungeunbae:
create_directory() is generic enough to be in qcore.


sungeunbae:
Fair point. Using setup_dir() instead. Pull requested withdrawn.

PR #107, opened 2019-04-05
deleted unused function load _params

ttgenius:
deleted unused function load _params
qcore.utils.update() function uses recursion, hence I did not move it inside the load_sim_params function


ttgenius:
updated

PR #108, opened 2019-04-09
QSW 1057

claudio525:


qcore/constants.py
```diff
@@ -175,14 +176,15 @@ class Components(ExtendedEnum):
...
-class State(ExtendedStrEnum):
+class Status(ExtendedStrEnum):
     """Job status on the HPC"""
 
     created = 1, "created"
     queued = 2, "queued"
     running = 3, "running"
     completed = 4, "completed"
     failed = 5, "failed"
+    unknown = 6, "unknown"
```
jasonmotha: Technically it should be error. But I guess this depends on what your other pull request looks like.

claudio525: This is for the scenario, when a task on the queue takes on a different status than R, PD (+CG in my workflow changes). Previously the status was just set to empty, when it wasn't one of those.
This was quite a common scenario, as tasks often briefly get set to CG.

PR #109, opened 2019-04-15
Update timeseries.py

lee-robin:
Added acc2txt and all2acc, all2acc uses acc2txt.

Basically copied vel2txt and all2txt and then changed vel 2 acc.

Feel free to actually change the function names if you think something else is more appropriate.

### DIFF
--- a/qcore/timeseries.py
+++ b/qcore/timeseries.py
@@ -347,47 +347,33 @@ class LFSeis:
             dt = self.dt
         return vel2acc3d(self.vel(station, dt=dt), dt)
 
-    def vel2txt(self, station, prefix='./', title='', dt=None):
+    def vel2txt(self, station, prefix='./', title='', dt=None, acc=False):
         """
         Creates standard EMOD3D text files for the station.
         """
         if dt is None:
             dt = self.dt
-        for i, c in enumerate(self.vel(station, dt=dt).T):
+        if acc:
+            f = self.acc
+        else:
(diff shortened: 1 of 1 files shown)


qcore/timeseries.py
```diff
@@ -368,6 +378,17 @@ def all2txt(self, prefix='./', dt=None):
         for s in self.stations.name:
             self.vel2txt(s, prefix=prefix, title=prefix, dt=dt)
 
+    def all2acc(self, prefix='./', dt=None):
```
viktor76525: all2txt does not imply acc or vel.
therefore instead of copying function differing only by vel2txt, acc2txt, add option that sets func = vel2txt or acc2txt based on parameter

qcore/timeseries.py
```diff
@@ -357,6 +357,16 @@ def vel2txt(self, station, prefix='./', title='', dt=None):
             seis2txt(c, dt, prefix, station, self.COMP_NAME[i],
                      start_sec=self.T_START, title=title)
 
+    def acc2txt(self, station, prefix='./', title='', dt=None):
+        """
+        Creates standard EMOD3D text files for the station (acc).
+        """
+        if dt is None:
+            dt = self.dt
+        for i, c in enumerate(self.acc(station, dt=dt).T):
```
viktor76525: again, only differing in self.acc vs self.vel? parameter rather than copy and paste

PR #114, opened 2019-04-29
Change to have and use vm_params.yaml as a constant, rather than a

jpa198:
repeated string

### DIFF
--- a/qcore/validate_vm.py
+++ b/qcore/validate_vm.py
@@ -18,7 +18,12 @@ import os
 import sys
 import argparse
 from qcore.utils import load_yaml
-from qcore.constants import VM_PARAMS_FILE_NAME
+from qcore.constants import VM_PARAMS_FILE_NAME, VMParams
+from qcore.simulation_structure import (
+    get_VM_file,
+    get_fault_VM_dir,
+    verify_VM_files_exist,
+)
 
 DEM_PATH = "/nesi/project/nesi00213/opt/Velocity-Model/Data/DEM/NZ_DEM_HD.in"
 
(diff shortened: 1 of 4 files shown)


qcore/validate_vm.py
```diff
@@ -51,12 +52,12 @@ def vmfile(filename):
     for fixed_name in vm.values():
         if not os.path.exists(fixed_name):
             return False, 'VM file not found: %s' % (fixed_name)
-    if not os.path.exists(vmfile('vm_params.yaml')):
+    if not os.path.exists(vmfile(VM_PARAMS_FILE_NAME)):
```
jasonmotha: While we are here it probably would be a good idea to move more of this logic to structure.py 
i.e. the VM file function too.

jasonmotha: Can also move the writing to this same location too.

qcore/validate_vm.py
```diff
@@ -18,129 +18,185 @@
...
     if numpy:
-        # check first zx slice (y = 0)
-        smin = np.min(np.fromfile(vm['s'], dtype = '<f%d' % (SIZE_FLOAT), \
-                count = vm_conf['nz'] * vm_conf['nx']))
-        pmin = np.min(np.fromfile(vm['p'], dtype = '<f%d' % (SIZE_FLOAT), \
-                count = vm_conf['nz'] * vm_conf['nx']))
-        dmin = np.min(np.fromfile(vm['d'], dtype = '<f%d' % (SIZE_FLOAT), \
-                count = vm_conf['nz'] * vm_conf['nx']))
+        mins = []
+        for file_name in vm.values:
```
claudio525: Needs to be vm.values()

PR #116, opened 2019-05-01
Remove cybershake cfg

jpa198:


jpa198:
Can't be merged until workflow pull request #162 is in

PR #131, opened 2019-05-30
Implementation of Source, Site and GMData + some fixes.

claudio525:


qcore/nhm.py
```diff
@@ -0,0 +1,119 @@
...
+                Row 11: MwMedian , RecurIntMedian  (yr)
+                Row 12: Num Locations on Fault Surface
+                Row 13+: Location Coordinates (Long, Lat)
+
+        Returns
+        -------
+        NHMInfo
+        """
+        # Not sure of a tidier of doing this (without adding all these to the init..)
+        info = NHMInfo()
```
sungeunbae: Just curious. Can't you put all these in the init, and then, return NHMInfo(x,x,x,x,x,)?

qcore/nhm.py
```diff
@@ -0,0 +1,119 @@
...
+        info.mw_median, info.recur_int_median = cls.read_row(rows[10])
+        info.n_locations = int(rows[11].strip())
+
+        # Read the locations
+        info.locations = np.asarray([cls.read_row(row) for row in rows[12:]])
+
+        return info
+
+    @staticmethod
+    def read_row(row: str, sep: str = None, convert: bool = True):
```
sungeunbae: Can we just embed this function? No one except from_nhm_section() (or __init__ we can do all this inside it) will need to call read_row()..staticmethod seems a bit too much.

qcore/nhm.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+    # Read the file and skip the first 15 lines (format info)
+    with open(nhm_file, "r") as f:
+        rows = f.readlines()[15:]
+
+    cur_rows, result = [], []
+    for row in rows:
+        # Empty row, i.e. separates faults
+        if len(row.strip()) == 0:
+            result.append(NHMInfo.from_nhm_section(cur_rows))
```
sungeunbae: Oh, I can see why you didn't want to put everything in the __init__ ()  NHMInfo(cur_rows) looks kinda bad.. but from_nhm_section() can still parse cur_rows and return NHMInfo(xxxxx.....)?

PR #135, opened 2019-06-05
HF seed now has a default value set in constants

jpa198:
This value is used as the default in all locations where the hf seed is passed


qcore/constants.py
```diff
@@ -7,6 +7,7 @@
 
 HF_DEFAULT_NCORES = 80  # 1 node, hyperthreading
 HF_DEFAULT_VERSION = "run_hf_mpi"
+HF_DEFAULT_SEED = 0
```
jasonmotha: Include a comment to say what the behaviour is.

PR #145, opened 2019-07-11
simplified nhm.py

viktor76525:


viktor76525:
version of nhm taken from seistech

PR #147, opened 2019-07-25
b'error' to 'error' as shared.exe now return str instead of byte

ttgenius:
Change b'error' to 'error' as shared.exe now return str instead of byte


ttgenius:
> Good point! Will you scan if there are other cases (across all other repos) where shared.exe() is used and assumed b"xxx" return value?

Searched through all the git repos, one place in qcore found still assuming b'xxx'. Have updated it as well. Thanks!

PR #157, opened 2019-08-22
Added crosstrack distance with optional shortcuts

jpa198:
Used for R_x
Other formatting


jasonmotha:
DISC: I have no idea how to verify the math. Do you have a reference?

jpa198:
> DISC: I have no idea how to verify the math. Do you have a reference?

https://www.movable-type.co.uk/scripts/latlong.html#cross-track Is the source I used.

jasonmotha:
Looks good.

PR #158, opened 2019-08-23
Metadata

ttgenius:
To run with metadata branch of slurm_gm_workflow.
Added class method to check substring because need to load submit_time_1,submit_time2 etc in addition to submit_time

### DIFF
--- a/qcore/constants.py
+++ b/qcore/constants.py
@@ -60,6 +60,10 @@ class ExtendedEnum(Enum):
     def has_value(cls, value):
         return any(value == item.value for item in cls)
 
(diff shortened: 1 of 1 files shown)


qcore/constants.py
```diff
@@ -253,6 +259,7 @@ class Status(ExtendedStrEnum):
     completed = 4, "completed"
     failed = 5, "failed"
     unknown = 6, "unknown"
+    cancelled = 7, "cancelled"
```
jasonmotha: When does the cancelled state occur in management?

PR #159, opened 2019-09-04
Gmsimviz fixes

viktor76525:
Hikurangi seems to be running good now.

### DIFF
--- a/qcore/gmt.py
+++ b/qcore/gmt.py
@@ -1135,10 +1135,11 @@ def table2grd(
                 for _ in range(header):
                     tf.readline()
                 # assert added to catch eg: first line = '\n'
-                assert len(map(float, tf.readline().split()[:2])) == 2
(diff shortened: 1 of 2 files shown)


qcore/gmt.py
```diff
@@ -2421,6 +2426,8 @@ def clip(self, path=None, is_file=False, invert=False, n=None):
             cmd = [GMT, "psclip", "-J", "-R", "-K", "-O", self.z]
             if invert:
                 cmd.append("-N")
+            if self.p:
```
sungeunbae: Can you explain why perspective option is handled in this function, and how this is related to the problem you are trying to solve?

viktor76525: in __init__
# perspective mode, 3D mode default
self.p = False

It's a bool that determines if the plot has perspective.

PR #162, opened 2019-09-20
without -R, assumes grid region==plot region

viktor76525:
bug where adding contours would stretch contour map across the whole map even if they were just for a subregion

### DIFF
--- a/qcore/gmt.py
+++ b/qcore/gmt.py
@@ -55,7 +55,7 @@ if not os.path.isdir(GMT_DATA):
     try:
         GMT_DATA = os.environ["GMT_DATA"]
     except KeyError:
-        # GMT_DATA files unavailable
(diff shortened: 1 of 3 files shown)


qcore/gmt.py
```diff
@@ -51,6 +51,12 @@
 GMT52_POS = {"map": "g", "plot": "x", "norm": "n", "rel": "j", "rel_out": "J"}
 
 GMT_DATA = qconfig["GMT_DATA"]
+if not os.path.isdir(GMT_DATA):
+    try:
+        GMT_DATA = os.environ["GMT_DATA"]
+    except KeyError:
+        # GMT_DATA files unavailable
+        pass
```
sungeunbae: pass? not sure what the expected behaviour is?

PR #164, opened 2019-10-03
fix from gmsimviz

viktor76525:


qcore/xyts.py
```diff
@@ -50,10 +50,12 @@ def __init__(self, xyts_path, meta_only=False):
         self.x0, self.y0, self.z0, self.t0, self.nx, self.ny, self.nz, self.nt = np.fromfile(
             xytf, dtype="%si4" % (endian), count=8
         )
-        self.dx, self.dy, self.hh, self.dt, self.mrot, self.mlat, self.mlon = np.fromfile(
+        self.dx, self.dy, self.hh, dt, self.mrot, self.mlat, self.mlon = np.fromfile(
             xytf, dtype="%sf4" % (endian), count=7
         )
         xytf.close()
+        # dt is sensitive to float error
+        self.dt = np.around(dt, decimals=4)
```
sungeunbae: Can you add why this step is needed by adding an example?

viktor76525: floating point error is instead of 0.2, you might get 0.1999999 or 0.2000001, therefore the number is incorrect and I have commented on this. what else should I add?

sungeunbae: Exactly what you explained here can be added to the comment to help the readability - regard the reader of your code as dumb and stupid - an example supplementing a statement will enlighten an ageing soul like me greatly :)

viktor76525: new comment

viktor76525: git hub is not showing this as outdated? it's been changed

PR #165, opened 2019-10-03
Must be converted to list in python3

jpa198:


jpa198:
> What code uses this? Wouldn't a pandas read_csv be easier and more appropriate?

Just visualisation I think. 
This is intended to be a get it from not working to working patch, not a proper fix.

PR #167, opened 2019-11-01
Timeshift changes

jpa198:
Mostly just changes to LFSeis and adding a constant for EMOD3D v304
Blacked
Breaking change for LFSeis though, requires e3d.par symlink in OutBin

### DIFF
--- a/qcore/utils.py
+++ b/qcore/utils.py
@@ -17,6 +17,7 @@ class DotDictify(dict):
     Construct an dictionary object whose values can also be accessed by 'dot'
     eg. d.k; d.k1.k2
     """
+
     MARKER = object()
 
     def __init__(self, value=None):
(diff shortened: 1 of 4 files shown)


qcore/utils.py
```diff
@@ -200,7 +209,28 @@ def load_py_cfg(f_path):
...
     with open(f_path) as f:
-        module = imp.load_module('params', f, f_path, ('.py', 'r', imp.PY_SOURCE))
+        module = imp.load_module("params", f, f_path, (".py", "r", imp.PY_SOURCE))
         cfg_dict = module.__dict__
 
     return cfg_dict
+
+
+def compare_versions(version1, version2):
+    parts1 = version1.split(".")
```
jasonmotha: Comment to state what each return code means.

Actually, an enum would be pretty cool.

PR #170, opened 2019-12-04
Adv im

shot0829:


qcore/constants.py
```diff
@@ -147,7 +147,7 @@ class ProcessType(ExtendedStrEnum):
         False,
         False,
         "time python $IMPATH/calculate_ims.py {sim_dir}/BB/Acc/BB.bin b -o {sim_dir}/IM_calc/ -np {np} -i "
-        "{sim_name} -r {fault_name} -c {component} -t s {extended} {simple}",
+        "{sim_name} -r {fault_name} -t s {component} {extended} {simple} {advanced_IM}",
```
jasonmotha: Wouldn't this run it with the advanced im flags?

qcore/timeseries.py
```diff
@@ -293,12 +293,16 @@ def __init__(self, outbin):
...
         self.seis = sorted(glob(os.path.join(outbin, "*seis-*.e3d")))
+        #try load e3d.par at the same directory first
+        #otherwise try look for one folder above
         self.e3dpar = os.path.join(outbin, "e3d.par")
         if not os.path.isfile(self.e3dpar):
-            raise ValueError(
-                "The given directory does not contain an e3d.par file. "
-                "Either move or create a symlink to the correct file please."
-            )
+            self.e3dpar = os.path.join(outbin, "../e3d.par")
```
jpa198: In an ideal world I don't think we should be looking outside the given directory, but this is ok as a "It's gets the job done" fix I think. Maybe add a flag to the user that it had to look outside?

qcore/constants.py
shot0829: the default value of the advanced_IM is empty.
so one of {component} and {advanced_IM} will be empty while the other have value.

jpa198:
Changelog!

PR #171, opened 2019-12-10
Unperturbated empiricals

jpa198:


qcore/simulation_structure.py
```diff
@@ -2,9 +2,21 @@
 Gives access to the folder structure of the cybershake directory
 """
 import os
+from pathlib import Path
+from typing import Union
 
 import qcore.constants as const
 
+
+def path_joiner(head, *tails):
```
jasonmotha: so functionally equivalent to os.path.join ?

I presume created because it's easier to convert them rather than changing the syntax?

jasonmotha: Can you do this non-recursively? Recursion generally isn't the best coding practice.

jasonmotha: Personally, I don't see a reason for this convinience function, it's as much effort to call this function as it is to append to the path?

jpa198: os.path.join(Path, Path) returns a string, so this is here to allow for both strings and pathlib.Path objects to be used, and a Path is returned if at least one element was.
Will change to non-recursive though

PR #179, opened 2020-03-16
Emp auto

sungeunbae:
Necessary changes for automating empirical calculation

### DIFF
--- a/qcore/simulation_structure.py
+++ b/qcore/simulation_structure.py
@@ -7,6 +7,7 @@ import qcore.constants as const
 
 
 def get_fault_from_realisation(realisation):
+    realisation = os.path.basename(realisation) #if realisation is a fullpath
     return realisation.split("_")[0]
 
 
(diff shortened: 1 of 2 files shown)


qcore/simulation_structure.py
```diff
@@ -7,6 +7,7 @@
 
 
 def get_fault_from_realisation(realisation):
+    realisation = os.path.basename(realisation) #if realisation is a fullpath
```
sungeunbae: Strictly speaking this is no longer necessary. 
When I started this task, some routine called this function with a full path to the realisation (that contained extra underscores), which caused the the weird directory structure like  /nesi/nobackup/nesi00213/RunFolder/EndToEndTest/tests/tmp_20200311_035853/Runs/Hossack/Hossack_HYP01-10_S1244/IM_calc/**rrup_/nesi/nobackup/nesi00213/**  where it should be rrup_Hossack.csv

PR #186, opened 2020-06-18
Siteamp

jasonmotha:
Added a new siteamp model. Most of the changes below are black but two functions have been extracted from CB14 and then a couple of implementation functions added. Where possible I have made the function signatures similar to maintain compatibility and to make swapping models easier.


jasonmotha:
> No idea whats going on there but seems legit. Did you write the new code or is this adding in an existing implementation?

Existing implementation. Ripped out a few lines of python from a file and shoehorned it into what we had.

PR #199, opened 2020-09-08
Siteamp

jasonmotha:


lee-robin (changes requested):
One comment below regarding another feature. Would prob like to test quickly.
Edit: I think my comment didn't go through and I accidentally did a duplicate review. Basically everything looks good but I was commenting on line 358 where it does the amp_bandpass it still seems to taper to unity from fhightop to fmax and beyond. Wasn't sure if it was meant to done for this PR but thought I'd comment and see anyway?

lee-robin (changes requested):
One comment below regarding another feature. Would prob like to test quickly.

qcore/siteamp_models.py
```diff
@@ -321,24 +325,42 @@ def ba18_amp(
...
+    vsite, freqs = ba_18_site_response_factor(vs, pga, vpga)
 
     amp = np.exp(vsite - ref)
-    # interpolation function expects frequencies in descending order
-    ampv, ftfreq = interpolate_frequency(freqs[::-1], amp[::-1], dt, n)
+    ftfreq = get_ft_freq(dt, n)
 
-    ampf = amp_bandpass(ampv, fhightop, fmax, fmidbot, fmin, ftfreq)
+    ampi = np.interp(ftfreq, freqs, amp)
+    # ampfi = amp_bandpass(ampi, fhightop, fmax, fmidbot, fmin, ftfreq)
```
lee-robin, line 358: I think removing the high frequency taper this way doesn't quite work because:
1. Now it doesn't apply the low frequency taper (using fmidbot and fmin).
2. The function amp_bandpass actually adds in another entry into the array at the start (i.e. line 174 makes a vector of ones that is one entry longer, and then at line 176 the vector gets populated starting from index 1). This makes ampi one entry shorter than ampfi.

My suggestion is to uncomment this and change the returned variable back to ampfi. For this line, we just make fhightop =999 and fmax=1000 or something like that.

lee-robin (changes requested):
One comment, hopefully I've done it right this time.

qcore/siteamp_models.py
lee-robin: Just in case there's an misinterpretation, we should do fhightop=999 and fmax=1000 for line 358 only (i.e. only for BA18 site amp). Keep the amp_bandpass general because CB14 will still need it the way it is.

lee-robin (approved):

PR #214, opened 2021-02-16
updates for components for validation PR

jasonmotha:


qcore/formats.py
```diff
@@ -49,7 +49,7 @@ def load_im_file_pd(imcsv, all_ims=False, comp=None):
     df = pd.read_csv(imcsv, index_col=[0, 1])
 
     if not all_ims:
-        df = df[[im for im in df.columns if (len(im) < 12)]]
+        df = df[[im for im in df.columns if (len(im) < 15)]]
```
claudio525, line 52: This seems quite arbitrary? Why not define a set of standard IMs? 
e.g.
```
STANDARD_IMs = ["PGA", "PGV", "pSA"]
df = df[[im for im in df.columns if im.split("_")[0] in STANDARD_IMS]]
```
Still not perfect, but at least it defines what standard IMs are?

qcore/constants.py
```diff
@@ -353,6 +353,7 @@ class Components(ExtendedStrEnum):
     crotd50 = 4, "rotd50"
     crotd100 = 5, "rotd100"
     crotd100_50 = 6, "rotd100_50"
+    cnorm = 7, "norm"
 
     @staticmethod
     def get_comps_to_calc_and_store(arg_comps: List[str]):
```
jpa198, line 359: Is the logic for get_comps_to_calc_and_store still valid with the norm component?
Sounds similar to geom, so probably ok?

qcore/formats.py
jasonmotha: The reason it's set to an arbitrary length is that it's the filter for filtering out extended pSA periods. It's not the standard set of IMs. I guess a comment would help that be clarified.

jasonmotha: There are comments on the function header  - defines what the argument does.  Should I put a comment on the line too?

Or do you have a good idea to determine if it's an extended pSA period?

qcore/constants.py
jasonmotha: Yep should be fine. Norm is generated from Adv_IM so it's not a 'basic' component.

qcore/formats.py
claudio525: What qualifies as extended period? They are longer because they have lots of decimal places?

claudio525: I did see the docstring, however from that it wasn't clear to me, that this is specifically to target extended pSA periods.

claudio525: Do we have list of the standard pSA periods? If we do, then why not just check if im is not pSA or in the list of standard pSA periods?

jpa198: Standard pSAs are listed in the root_params.yaml files. Extended are generated from a linear spaced log scale.
We should be able to generate a superset of standard pSAs, the main issue being inconsistent floating point representation (I've had issues with trailing 0s before)

jasonmotha: extended pSA period is the 100 numbers logspaced between 0.01 and 10. Because they are on a computed distribution between 0.01 and 10 they have lots of decimal places.

We do have a list of standard pSA periods - but that is dependant on which gmsim version you run (maintaining identical functionality for old runs) and doesn't preclude the user adding an extra pSA period when running. The list of extended periods is less likely to change - but it's much harder to compare and Vahid did request a custom list recently.

jasonmotha: For a given IM_file it is unknown what pSA periods will be in the file without looking at the header.

shot0829: the default value is here:
https://github.com/ucgmsim/qcore/blob/ba419732e77083bf0842ab531d7b8509c5d82bcf/qcore/constants.py#L50
and with the nature of that function the length of the float64 is usually longer than what the standard pSA we run.
Hence filtering it with len of the IM name (e.g pSA_0.1 vs pSA_0.010722672220103232 )

shot0829 (approved):

jpa198 (dismissed):

qcore/formats.py
jpa198: And you can't just use EXT_PERIOD as the non basic periods, as there is some overlap with the basics (0.01, 0.1, 1, 10) (I think)

shot0829: @jpa198 correct, that's why we use https://github.com/ucgmsim/IM_calculation/blob/07b21f099272d6707daa95bfad31f23d99a3a37e/IM_calculation/IM/im_calculation.py#L527

claudio525: I guess if there isn't a better option, then the length is the best we can do. Maybe modify the comment/docstring slightly to make this more obvious?

jpa198 (approved):

claudio525 (approved):

shot0829 (approved):

PR #215, opened 2021-02-19
added vm_params.yaml location into sim structure

shot0829:


jpa198 (approved):
Required separate from the get_vm_params_path as it's used in pre-pro without access to the cs_root or event name

jasonmotha (approved):

PR #216, opened 2021-02-23
extended extended enum to be able to return a list of the names

jasonmotha:
Just double checking that there isn't a built-in way to do this otherwise this is perfect.


jpa198 (approved):
Seems ok, where is it going to be used though?

jasonmotha:
I'm just about to use in a script in empirical engine. I'm pretty sure we do similar functionality messily without wrappers elsewhere.

claudio525 (approved):

PR #220, opened 2021-03-18
Moves IM calc enums to constants

jpa198:


claudio525 (approved):

shot0829 (approved):

jasonmotha:
I'd argue that they don't belong in constants as they aren't used outside of IM_calc submission and aren't likely to be. @jpa198

PR #222, opened 2021-03-23
Add h1 and h2 to the components list

jpa198:
Add function to return the currently available basic components
Refactor read_ascii_folder to load arbitrary basic components, including h1, h2
Formatting


qcore/constants.py
```diff
@@ -396,6 +398,16 @@ def component_sorter(x):
 
         return components_to_get, components_to_store
 
+    @staticmethod
+    def get_basic_components():
+        return [
+            Components.c090,
+            Components.c000,
+            Components.cver,
+            Components.ch1,
```
jasonmotha, line 407: I'm not convinced that ch1 is a basic component?

jasonmotha (commented):
Generally happy with this side of things.

qcore/constants.py
jpa198: The advanced ones are all derived, so it definitely doesn't fit with them

jasonmotha (approved):

claudio525 (approved):

qcore/constants.py
shot0829: If i'm not wrong, h1 h2 from definition has to be calculated after the waveform has been recorded?
its just happens to be in ascii formatting. 
If the word 'basic' become confusing, we can change the naming.
I personally doesnt have too much issue with this, but i can see why jason questioned it at the first place

shot0829: however, if we actually are considering the two as basic components, wouldn't it be better to insert them into the list instead of appending? just for classification.
like the first few is always basic components, and then comes the advanced
so we wont have random sequences of basic and advanced components.

jasonmotha: h1, h2 are the raw waveforms recorded at station. 000, 090 are rotated to the cardinal directions - North and East

It doesn't really matter the order of them, I generally prefer adding them in order of history.

shot0829 (approved):

PR #224, opened 2021-04-18
add in opensees path to maui config

shot0829:


qcore/configs/machine_maui.json
```diff
@@ -1,6 +1,7 @@
 {
     "GMT_DATA" : "/nesi/project/nesi00213/PlottingData",
     "tools_dir" : "/nesi/project/nesi00213/opt/maui/hybrid_sim_tools/",
+    "OpenSees" : "/nesi/project/nesi00213/opt/maui/software/OpenSees/tmp/OpenSees",
```
jasonmotha, line 4: Pointing to a tmp folder doesn't seem right?

jasonmotha (changes requested):

qcore/binary_version.py
```diff
@@ -27,9 +27,12 @@ def get_lf_bin(version, bin_loc=default_bin_loc):
...
-    return path.join(bin_loc,  GENSLIP_NAME_STR.format(version))
+    return path.join(bin_loc, GENSLIP_NAME_STR.format(version))
 
 
 def get_unversioned_bin(bin_name, bin_loc=default_bin_loc):
     return path.join(bin_loc, bin_name)
 
+
+def get_opensees_bin():
+    return qconfig["OpenSees"]
```
jpa198, line 38: As OpenSees is a tool we build for our own use, this should be joined to bin_loc, and the machine config changed to not have the full path.
If a full path is required (Say for an external build), then the machine config can have the full path, and join will discard the bin_loc.

jpa198 (changes requested):

qcore/binary_version.py
shot0829: as opensees is a seperate external source code, i dont think putting it under the emod3d directory makes too much sense?
Its just like gmt, where we have a seperate opt directroy for it, as well as its dependencies.

qcore/timeseries.py
```diff
@@ -897,3 +897,11 @@ def acc(self, station, comp=Ellipsis):
             np.frombuffer(o, dtype=np.int32).astype(np.float32).reshape(-1, 3)[:, comp]
             / self.stations.scale[i]
         )
+
+def get_observed_stations(observed_data_folder):
```
jasonmotha, line 901: doc string should say what the function returns.

jasonmotha (changes requested):

qcore/binary_version.py
jpa198: Hmm, ok, I think this will be fine. I have some thoughts on how non emod3d repo binaries should be handled, but that's beyond the scope of this PR

qcore/timeseries.py
jpa198: @shot0829 The doc string for this still needs to be improved. A statement of what it does, and what the return value is would be good

qcore/timeseries.py
```diff
@@ -898,10 +898,12 @@ def acc(self, station, comp=Ellipsis):
...
+
 def get_observed_stations(observed_data_folder):
     """
     observed_data_folder: path to the record folder, e.g. observed/events/vol1/data/accBB/
     """
-    search_path = os.path.abspath(os.path.join(input_path, "*"))
+    search_path = os.path.abspath(os.path.join(observed_data_folder, "*"))
     files = glob(search_path)
-    station_names = set([ os.path.splitext(os.path.basename(x))[0] for x in files]
+    station_names = set([os.path.splitext(os.path.basename(x))[0] for x in files])
```
jasonmotha, line 908: `for filename in files`

jpa198 (approved):

qcore/timeseries.py
```diff
@@ -910,5 +910,7 @@ def get_observed_stations(observed_data_folder):
     """
     search_path = os.path.abspath(os.path.join(observed_data_folder, "*"))
     files = glob(search_path)
-    station_names = set([os.path.splitext(os.path.basename(x))[0] for x in files])
+    station_names = set(
+        [os.path.splitext(os.path.basename(filename))[0] for filename in files]
+    )
```
jpa198, line 915: Instead of set([list comprehension]) you can do {list comprehension} to get the set directly

jpa198 (changes requested):

qcore/timeseries.py
```diff
@@ -897,3 +897,20 @@ def acc(self, station, comp=Ellipsis):
...
+    :param observed_data_folder: path to the record folder, e.g. observed/events/vol1/data/accBB/
+    :type observed_data_folder: str/os.path
+    :return: list of unique station names
+    :rtype: list[str]
+    """
+    search_path = os.path.abspath(os.path.join(observed_data_folder, "*"))
+    files = glob(search_path)
+    station_names = set(
+        [os.path.splitext(os.path.basename(filename))[0] for filename in files]
+    )
```
jpa198, line 915: Possible github glitch, so I'll put this in again:
This should use a set comprehension instead of converting a list comprehension

jpa198 (changes requested):

jasonmotha (approved):

jpa198 (approved):

PR #225, opened 2021-04-23
srf cnr not in perimeter bugfix

viktor76525:


jpa198 (approved):

qcore/srf.py
```diff
@@ -836,13 +836,12 @@ def get_perimeter(srf_file, depth=True, plot=False):
...
-            c1 = np.argwhere(np.minimum.reduce(perimeters[-1] == points[0], axis=1))[0][
-                0
-            ]
-            c2 = np.argwhere(
-                np.minimum.reduce(perimeters[-1] == points[nstk - 1], axis=1)
-            )[0][0]
+            # try to find top edges in perimeter
+            # closest point in case corner not in perimeter
+            c1 = perimeters[-1] - points[0]
+            c1 = np.argmin(np.sqrt(np.einsum("ij,ij->i", c1, c1)))
```
jpa198, line 842: After having a look, this just seems to be taking the sum of the squares of the lon, lats differences? 
Given the difference in length of a lat and a lon near NZ, this should be handled by the geo.get_distances function
Performance shouldn't be an issue

jpa198 (changes requested):

qcore/srf.py
shot0829: this seems to be called by visualization/gmt/plot_srf_map.py only. and I'm not sure stuffs in there consider earth curvature while geo.get_distance does?

jpa198: Even if plot_srf_map doesn't use curvature, this line gets the nearest point to the given location, so the size of lats and lons does matter.
Around NZ 1 degree of lon is ~81km, while 1 degree of lat is ~111km

jpa198:
@viktor76525

qcore/srf.py
viktor76525: It doesn't matter. Usually distance will be 0. Sometimes distance may be a couple of metres if the roughness staggered the points around a bit (subfaults).
It wouldn't make sense to worry about curvature, only to worry about depth. 3d cartesian points will pretty much guarantee finding the correct closest point.
Dealing with curvature would be high precision but absolutely no accuracy, the next step is depth.

jpa198: Cool, you should at least add that as a comment otherwise it looks like a bug.
It can still be simplified by removing the sqrt of a value if you're taking the argmin of it after. It'll also be easier to understand if you refactor the einsum call to the sum of the squares
The biggest upside to using geo.get_distances is the readability of the code. But if you can make this line as readable then that point is moot.

jpa198: In lieu of replacing the einsum call, can you add another comment to explain what it does? 
Once that's done I think this is ok to go in

viktor76525: put something in

jpa198 (approved):

shot0829 (approved):

PR #226, opened 2021-05-12
Added EAS component to handle Euclidean distance of FAS 000 and 090

sungeunbae:
Related to https://github.com/ucgmsim/IM_calculation/pull/105 
Added a separate component EAS, included as an advanced_component.
This will make the selection of EAS component to enforce the horizontal components (000, 090) to be added to the comps_to_calculate list.


qcore/constants.py
```diff
@@ -387,7 +388,7 @@ def component_sorter(x):
         # The horizontal components generated by simulation workflow steps
         horizontal_components = {Components.c090, Components.c000}
         basic_components = set(Components.get_basic_components())
-        advanced_components = set(list(Components)[3:8])
+        advanced_components = set(list(Components)[3:8]+[Components.ceas])
```
jasonmotha, line 391: instead of adding it as a list, I think that reordering it so that H1 & H2 are after ver?

Thoughts @jpa198

shot0829: i think this come downs to if we want the order to be chronicle or functional.
Last time this topic was brought up we decided chronicle makes it easier to trace.

sungeunbae: I was slightly worried about the change of order breaking up existing functionalities - lots of stuff depend on qcore..

jpa198: I think it should be reordered. {090, 000, ver}, {H1, H2}, {Derived}. Or something like that.
The order of the components shouldn't matter though, as we shouldn't be referencing them by index anywhere (even inside the enum), only by name. 
I'm in favour of removing the integer values from the entries too, while we're on the subject. Or at least replacing them with calls to auto().
Casting to a list and using indices here was just a convenience so we didn't have to explicitly list each component in its relevant type.

sungeunbae: @jpa198  Your comment about the order not being important is assuring. I rearranged the order. I didn't go as far as removing the integer values - it sounds like a lot more than originally scoped, and needs a bit more discussion.

jpa198: This is bit of a scream test, as while I can't think of anywhere using the values, there may still be somewhere using them. That's on the code writer using the values instead of the names though. Fair point on the removal though.

shot0829 (approved):

jasonmotha (approved):

PR #227, opened 2021-05-14
nhm.py writing / header functionality

jasonmotha:


qcore/nhm.py
```diff
@@ -98,3 +117,36 @@ def load_nhm(nhm_path: str, skiprows: int=15):
         faults[nhm_fault.name] = nhm_fault
 
     return faults
+
+
+def write_nhm_section(
```
claudio525, line 122: Tiny docstring and maybe name it `write_fault_section` or `write_nhm_fault`?

claudio525 (approved):

jasonmotha:
Note for me: extend NHM class with sample / file writing in obj.

jpa198 (approved):

jasonmotha:
This is a provisional PR. The only minor thing that will change is the SI Mw scaling relationship. 

The Mw scaling relationships aren't in pre-processing yet but will be part of another PR.

qcore/nhm.py
```diff
@@ -32,50 +60,129 @@ class NHMFault:
...
-                Row 11: MwMedian , RecurIntMedian  (yr)
-                Row 12: Num Locations on Fault Surface
-                Row 13+: Location Coordinates (Long, Lat)
-        """
-        rows = list(map(str.strip, entry.split("\n")))
+        kwargs of all the required parameters for a nhm fault
 
-        def str2floats(line):
-            return list(map(float, line.split()))
+        It is up to the caller to make sure all the required parameters are available
```
claudio525, line 70: Not sure I'm a fan of this, I think the init should take all the variables. Now it's very unclear what a NHM fault represents.

jasonmotha: I agree, I was being a little cheeky. Do you reckon filling the doc string is enough or should kwargs be unpacked?

claudio525: I think all required fields that make up a NHM fault should be specified directly. Don't have a problem with optionals in kwargs (I think). I know its a bit of effort, but will make future use much easier and clearer imo.

qcore/nhm.py
```diff
@@ -32,50 +60,205 @@ class NHMFault:
         fault surface trace (lon, lat)
     """
 
-    def __init__(self, entry: List[str]):
-        """Creates an NHMFault instance from the given NHM text.
+    def __init__(
```
jpa198, line 63: This can be replaced with member fields and the @dataclass decorator.
The main difference would be removing 2 of the 3 copies of each variable.
If we add other stuff to the initialiser then we'll need to put it back though.

jpa198 (changes requested):

qcore/nhm.py
```diff
@@ -32,50 +60,205 @@ class NHMFault:
...
-        self.coupling_coeff, self.coupling_coeff_sigma = str2floats(rows[9])
-        self.mw, self.recur_int_median = str2floats(rows[10])
-        self.trace = np.array(list(map(float, " ".join(rows[12:]).split()))).reshape((-1, 2))
         # TODO: add x y z fault plane data as in SRF info
         # TODO: add leonard mw function
 
+    def sample_2012(self):
+        """
+        Permutates the current NHM fault as per the OpenSHA implementation. This uses the same Mw scaling relations
+        in Stirling 2012
```
claudio525, line 143: as

qcore/nhm.py
```diff
@@ -32,50 +60,205 @@ class NHMFault:
...
+        Dtop is peturbated with a uniform distribution between min and max.
+        The remaining parameters are perturburbated with a truncated normal distribution (2 standard deviations)
+
+        :return: new NHM object with the perturbated parameters containing 0 in all sigma sections
+        """
+        rake = self.rake
```
claudio525, line 149: why not just use self.rake directly?

qcore/nhm.py
```diff
@@ -1,6 +1,34 @@
...
+Row 8: RupTopMean, RupTopMin RupTopMax  (km) 
+Row 9: SlipRateMean , SlipRateSigma (mm/yr) 
+Row 10: CouplingCoeff , CouplingCoeffSigma (mm/yr) 
+Row 11: MwMedian , RecurIntMedian  (yr) 
+Row 12: Num Locations on Fault Surface 
+Row 13+: Location Coordinates (Long, Lat) 
+"""
+
+
+MU = 3.0 * 10.0 ** 10.0
```
claudio525, line 27: Maybe add a comment or rename variable so that its clear what MU this is? Based on usage looks like moment magnitude mu?

qcore/nhm.py
```diff
@@ -1,20 +1,49 @@
...
+Row 11: MwMedian , RecurIntMedian  (yr) 
+Row 12: Num Locations on Fault Surface 
+Row 13+: Location Coordinates (Long, Lat) 
+"""
+
+# This mu is used to calculated Moment Rate
+MU = 3.0 * 10.0 ** 10.0
+
+
+def sample_trunc_norm_dist(mean, sigma, sigma_limit=2):
```
jpa198, line 31: This is a duplicate of the functions in srf_generation/source_parameter_generation/uncertainties/distributions.py
It's already here now. But could be imported from there with the same effect

qcore/nhm.py
```diff
@@ -29,53 +58,157 @@ class NHMFault:
...
+            mw_median, mw_sigma = mag_scaling.lw_2_mw_sigma_scaling_relation(
+                length, width, mw_scaling_rel, self.rake
+            )
+
+            if mw_perturbation:
+                mw = sample_trunc_norm_dist(mw_median, mw_sigma)
+            else:
+                mw = mw_median
+
+        moment = 10 ** (9.05 + 1.5 * mw)
```
jpa198, line 151: Where did this magnitude-moment relationship come from? It seems to differ from the mag2mom function in mag_scaling.py 10^(9.05+1.5*Mw) vs 10^(16.05+1.5*Mw) (I think)

jasonmotha: Correct, that one is incorrect (dyne-cms) while this one is moment metres. As part of the scaling relation PR, I've added comments / these functions.
image

jpa198 (approved):

claudio525 (approved):

PR #231, opened 2021-06-14
Add static method to generate IM object from str

jasonmotha:
Add static method that can generate an IM object from the output of the get_im_name function, as found in IM_calc csv headers


jasonmotha:
This doesn't look unuseful?

jpa198:
This looks good, and should make it easier to grab the IM dataclass value, I need to use it more...
I haven't had much of a look at this before, but does the post_init component line work? It looks like it tries to call from_str on what ever is in the variable name, instead of the Components enum

jpa198:
As an aside, should the ProcessType enum in constants get changed to mostly a dataclass? It feels too bloated to be an enum anymore. Not that it will change that much.

jpa198 (approved):

jasonmotha:
Yeah. That whole bit of the submission workflow needs a refactor. I think changing it to a dataclass puts a bandaid over the wreck.

jpa198 (approved):

shot0829 (approved):

PR #239, opened 2021-07-25
fix setup

viktor76525:


shot0829 (approved):

jasonmotha (commented):
Does qcore have a data folder? 

Or is that just for generating the package data?

viktor76525:
> Does qcore have a data folder?
> 
> Or is that just for generating the package data?

It's in the rest of setup.py, downloaded from the releases page.

jasonmotha (approved):

PR #244, opened 2021-08-19
Refactor validate_vm

jpa198:
…and workflow

Still needs kisti/tacc script updates


shot0829 (dismissed):

qcore/validate_vm.py
```diff
@@ -15,219 +15,238 @@
...
+        vm_dir / f"gridfile{vm_params_dict['sufx']}",
+        vm_dir / f"model_bounds{vm_params_dict['sufx']}",
+        vm_dir / f"model_coords{vm_params_dict['sufx']}",
+        vm_dir / f"model_params{vm_params_dict['sufx']}",
+    ]
+    for vm_file in all_files:
+        if not vm_file.exists():
+            errors.append(f"VM file not found: {vm_file}")
+
+    # 5, 6: binary file sizes
```
jasonmotha, line 123: Some of the numbering seems off in this file?

jasonmotha: Were you just keeping original numbering? It might be better to re-number the checks?

qcore/validate_vm.py
```diff
@@ -15,219 +15,238 @@
...
-    #        # not as important, can be re-created based on vm_params.py
-    #        pass
+def validate_vm_file(file_name, vm_size):
+    errors = []
+    size = file_name.stat().st_size
+    if size != vm_size * SIZE_FLOAT:
+        errors.append(
+            f"VM filesize for {file_name} expected: {vm_size * SIZE_FLOAT} found: {size}"
+        )
+    if (
```
jasonmotha, line 186: I'd add a comment here. Actually a docstring might be clearest. Would be good to state what is being tested here.

qcore/validate_vm.py
```diff
@@ -15,219 +15,238 @@
...
-    rc = 1
-    parser = argparse.ArgumentParser()
-    parser.add_argument("VM_dir", type=str, help="path the VM folder")
-    parser.add_argument("--srf", default=None, type=str, help="Path to the SRF to be used with this VM")
     args = parser.parse_args()
-    try:
-        success, message = validate_vm(
-            args.VM_dir, srf=args.srf
+
+    if args.subparser_name == "params":
```
jasonmotha, line 226: Is it worth having a CONST or enum or something for the subparser types?

jasonmotha (changes requested):
Looks good! I was originally thinking to have separate scripts but the subparsers work well.

Have you run this over a perturbation / .s file?

qcore/validate_vm.py
jpa198: Yeah, this is based off the original numbering. I should probably drop/re number it though

jpa198:
> Looks good! I was originally thinking to have separate scripts but the subparsers work well.
> 
> Have you run this over a perturbation / .s file?

Yeah, it seemed like a good time to try out sub parsers, especially given how small each individual script would end up being.
I ran it over a perturbation file and it seemed happy. Might need to turn the minimum value check to only first and last layer or something though. Will see if there are any performance issues.

qcore/validate_vm.py
```diff
@@ -15,219 +15,251 @@
...
+    )
+
+    vm_file_parser = sub_parser.add_parser(
+        NZVM_SUB_PARSER, help="Validate files generated by the NZVM"
+    )
+    vm_file_parser.add_argument("vm_dir", type=Path, help="path the VM folder")
+
+    pert_parser = sub_parser.add_parser(
+        SINGLE_FILE_SUB_PARSER,
+        help="Validate a single VM file for size and 0s. Primarily intended for perturbation files.",
```
sungeunbae, line 232: "file" sounds a little misleading, especially in presence of "files". While the option names are singular vs plural, what they actually do are very different. If this is only intended for sanity check including size and non-zero values, what about "sanity"? Could this be handling multiple files too?

sungeunbae (commented):
Just like to comment on option name - "file"

qcore/validate_vm.py
jpa198: I have been thinking about better names. I was thinking of 'NVZM' instead of 'files', as it's all based on the NZVM output. 
For our current use case I think leaving 'file' with its current name is ok, and only doing one file.
The entire script is a sanity check for VM files, just validating them and making sure they seem reasonable, so wouldn't be useful as an option name

jasonmotha:
One thing would be good is if we could do the DEM bounds check and the SRF within check as part of the parameter generation. But unsure if you have the corners file available for that part.

qcore/validate_vm.py
sungeunbae: nzvm sounds good. What about "quick" for "file"?

jasonmotha: Quick isn't necessarily a quick test. It's a test to check a single output file from the NZVM. So maybe a NZVM_binary or similar? 

Or we could call them: complete and perturbation as those are the use cases we are using them with. But they can be applied more generally than that.

sungeunbae: as it only checks the size and if it contains non-zero values, "quick"er than full-scale check

jpa198:
> One thing would be good is if we could do the DEM bounds check and the SRF within check as part of the parameter generation. But unsure if you have the corners file available for that part.
> How possible is this on a scale of 1 to possible?

Yeah, that should be possible. Off the top of my head parameter gen has to make its own version of the srf bounds, and can generate the equivalent of the corners file. Will have a better look now.

qcore/validate_vm.py
jpa198: It's not there to be a quick check, it's there to check a single file such as a perturbation or Qp/Qs file that isn't in the regular NZVM output, so I don't think the name 'quick' is appropriate. I'm not that keen on calling it NZVM_binary when the main use case is non NZVM binaries
My current preferred names are 'params', 'NZVM', 'file'

jasonmotha: I don't think the naming is a big issue as the scripts call will often be automated. I think most of the suggestions here have made sense; so happy with your judgement.

sungeunbae: I can settle on "params, NZVM, file".

jpa198:
I think we're good to go now? I did a test run on Maui/Mahuika using the manual vm gen and install scripts and it went smoothly. After I ironed out the bugs.

sungeunbae (approved):

qcore/validate_vm.py
```diff
@@ -15,219 +15,282 @@
...
+    :param polygon: A list of (lon, lat) tuples giving the corners of the VM
+    :param srf: A list of (lon, lat) tuples giving the corners of the VM. Not used if None
+    :return: A list of error messages resulting from this validation
+    """
+    errors = []
+
+    for lon, lat in polygon:
+        if lon < MIN_LON or lon > MAX_LON or lat < MIN_LAT or lat > MAX_LAT:
+            errors.append(f"VM extents not contained within NZVM DEM: {lon}, {lat}")
+    # Check SRF is within bounds of the VM if it is given
```
jasonmotha, line 194: Would be good to add a comment that this is checked twice.

It's probably a good idea to check it twice as there may be slight differences in the corners to the computed bounds? (is that right?)

jasonmotha (commented):
Looks good. Looks like the check to move the bounds to params weren't that bad.

qcore/validate_vm.py
jpa198: Checked twice, as in once in the python script, once in the slurm script?

jasonmotha: Once in the check_params once in the check_NZVM?

jpa198: So the expected corners are tested against the DEM and srf in check_params, while the actual corners are tested against the DEM in check_NVZM (In the current vm_params2vm the srf isn't checked, that's done later at the install step atm). I don't know too much about how NZVM works, but I do expect there to be minor differences between python calculated expected, and C calculated actual corners.

jpa198: I don't think it's necessary to add a comment that it gets checked twice? Who would the comment be for and why would it be useful to them?

jasonmotha: Oh yeah. I see how that's getting called now. 

Yeah, the comment itself wouldn't be useful. I was thinking from the perspective that it's clear to which tests are part of which file. Which then that becomes a Wiki documentation thing.
What do you think?

jpa198: So making a wiki page for vm validation? Sounds like a plan

jpa198:
https://wiki.canterbury.ac.nz/display/QuakeCore/Velocity+model+validation
Wiki page done. Anything else before we can merge?

jasonmotha (approved):
Perfect! I would have used a table to show what gets tested where but I think what you've done is clear enough.

jasonmotha:
Actually, while we are here can we rename it verify_vm?

sungeunbae:
Oh no - my home internet is very flaky today. Are you planning it to rename @jpa198 ?

jpa198:
We probably should. The follow on changes in other repos make it more annoying though. I think it's fine for now, and we try to bring it up earlier next time we edit vm validation/verification?

PR #247, opened 2021-09-01
Creating virtual environment for each test

sungeunbae:
Creates virtual environment at /tmp/qcore/#hash#, installs requirements.
Runs test, then removes after.

Docker images all removed from the server to release space.
Binaries are kept under /nesi/project/nesi00213/tools 

Note: For virtual environments, pytest and pytest-black should be included in requirements.txt for the auto-tests to run


Jenkinsfile
```diff
@@ -1,30 +1,47 @@
...
-		docker images sungeunbae/qcore-ubuntu-minimal -q
-		"""
-	    }
-	}
+        stage('Settin up env') {
+            steps {
+                echo "Start virtual environment"   
+                sh """ 
+                    pwd
+                    env
```
jasonmotha, line 9: If we're using these as debug output for the logs, it would be nice to wrap it with "Current working directory:" etc.

jasonmotha (changes requested):
@jpa198 Thoughts about compiling emod3d-tools as part of environment setup. Just because the binaries may change and hence break the tests?

jpa198:
> @jpa198 Thoughts about compiling emod3d-tools as part of environment setup. Just because the binaries may change and hence break the tests?

I think pulling EMOD3D and compiling every time we do a test is too much. I think the responsibility for updating the binaries should be on the EMOD3D repo. 
So that probably needs a jenkins file to build the tools on PR creation or merge.
EMOD3D is normally low traffic, so it shouldn't be an issue too often.

jasonmotha:
> > @jpa198 Thoughts about compiling emod3d-tools as part of environment setup. Just because the binaries may change and hence break the tests?
> 
> I think pulling EMOD3D and compiling every time we do a test is too much. I think the responsibility for updating the binaries should be on the EMOD3D repo.
> So that probably needs a jenkins file to build the tools on PR creation or merge.
> EMOD3D is normally low traffic, so it shouldn't be an issue too often.

Sweet. Agreed with all your thoughts.

Jenkinsfile
sungeunbae: You mean as a separate "stage"? Sure - can do.

sungeunbae:
> > > @jpa198 Thoughts about compiling emod3d-tools as part of environment setup. Just because the binaries may change and hence break the tests?
> > 
> > 
> > I think pulling EMOD3D and compiling every time we do a test is too much. I think the responsibility for updating the binaries should be on the EMOD3D repo.
> > So that probably needs a jenkins file to build the tools on PR creation or merge.
> > EMOD3D is normally low traffic, so it shouldn't be an issue too often.
> 
> Sweet. Agreed with all your thoughts.

I only managed to build a couple of tools on ec2 - had to temporarily install openmpi and FFTW3, and it crashed at Fractal3D. At the point, I had all the binaries needed for the test, so I didn't go further to fix the build issues.

Jenkinsfile
sungeunbae: Done in the latest patch

Jenkinsfile
```diff
@@ -1,19 +1,26 @@
...
                     pwd
                     env
+                """
+            }
+        }
+        stage('Settin up env') {
+            steps {
+                echo "Start virtual environment"   
+                sh """
+# Each stage needs custom setting done again. By default /biin/python is used.
```
jasonmotha, line 17: /bin/python*

Jenkinsfile
```diff
@@ -1,19 +1,26 @@
 pipeline {
     agent any 
     stages {
-        stage('Settin up env') {
+        stage('Checking env') {
             steps {
-                echo "Start virtual environment"   
-                sh """ 
+                echo "Check the environment"
```
jasonmotha, line 6: I'd still like some labelling on the lines below. Just so as a user you can tell what each path is from the logfile.

sungeunbae: I misunderstood what you meant. Reverted the previous change (separate check stage) and added more comments

jasonmotha (approved):
Cool. Looks good.

jpa198 (approved):

PR #249, opened 2021-09-09
update the exe function to be able to handle commands with shell=True

shot0829:


qcore/shared.py
```diff
@@ -57,16 +57,17 @@ def get_corners(model_params, gmt_format=False):
     return corners, cnr_str
 
 
-def non_blocking_exe(
-    cmd, debug=True, shell=False, stdout=True, stderr=True, **kwargs
-):
-    # always split for consistency
-    if type(cmd) == str:
+def non_blocking_exe(cmd, debug=True, shell=False, stdout=True, stderr=True, **kwargs):
+    # always split for consistency (But only when shell == False
```
jasonmotha, line 61: This comment needs revision.

jasonmotha (changes requested):
Can we get an example of where this was an issue?

qcore/shared.py
sungeunbae: > Can we get an example of where this was an issue?

I may be barking at the wrong tree, but vaguely remember when passing an argument surrounded by "", we used to have a problem.

shot0829: if the command is a list, instead of a string, the underlying behavior is different.
and this will cause issue if the shell/executable have a strict parsing-logic, in this case qsub with -v.
the values comes after -v and itself have to be parsed as one, and the easiest way to do so is execute the the command in one line with shell == True.

sungeunbae: So I was on the right track to guess that this issue is related to "". 

> qsub -v arg1="$var1/path",arg2='$foo',arg3=3 -otherflags script.sh

from https://wiki.canterbury.ac.nz/display/QuakeCore/PBS

sungeunbae (approved):

qcore/shared.py
jasonmotha: Ah yep. I think that argument should be handled differently to shell? As shell=True enables different functionality to the split/join argument handling.

To be honest splitting commands with .split() isn't the best practice in general - if we want to fix this properly it might be better to change the exe function to expect a list rather than a string?

jpa198: For this case I agree that the command should be submitted as a pre-split list, so that we can control which parts do and don't get split. Rather than changing how we process commands being submitted for execution

shot0829: the exe() can already handle cmd as a list.

jpa198: That's my point. Split the command earlier, not in here. IT means you can control how the split happens instead of just everything.
Making exe only accept lists is possibly a bit much, as it would result in a number of calls to split() further up the line.

jasonmotha: image
I was thinking of instead of creating the commands instead of long strings to create them as lists.
e.g.
[wget, -O, zip_download_path, DATA_DOWNLOAD_PATH]

sungeunbae: double instead of...

shot0829:
I just realized a way to acheive what i want from a level above. 
Doing a quick run to confirm my changes. 
This PR is less needed now. closing it.

qcore/shared.py
sungeunbae: I also remembered an occasion where plot_items.py -t "This is a title"  (I think it was spawned by compare_runs.py screwed up. https://github.com/ucgmsim/visualization/blob/master/compare/compare_runs.py )

I ended up with plot_items.py -t This_is_a_title.  I think it will be nice if a solution here can address that too.

PR #251, opened 2021-09-12
Ensure there is a minimum of 3 points along each edge of the srf

jpa198:
If the length was less than 1 km there were 0 points, resulting in an error from the Path


jasonmotha (approved):
Oooh, could this be influencing my bus error? -- maybe not but still interesting to consider.

shot0829 (approved):
ah ha , tiny sub-faults giving us more trouble.

jpa198:
> Oooh, could this be influencing my bus error? -- maybe not but still interesting to consider.

It shouldn't be, python should just raise the exception and call it a day

PR #253, opened 2021-09-20
Plot srf map bug

sungeunbae:
alphashape needs alpha parameter to be set (alpha=0 is a convex hull), but often we need a concave-hull. The question is how to determine the alpha parameter to give the best fit. Viktor originally found alpha=600 worked ok for SRF, but sometimes it doesn't return a legit Polygon object that will ensure the subsequent code works. 

This fix will, hopefully, find a reasonably good alpha value without impacting the performance too much.

### DIFF
--- a/qcore/srf.py
+++ b/qcore/srf.py
@@ -11,14 +11,17 @@ https://scec.usc.edu/scecpedia/Standard_Rupture_Format
 
 from math import ceil, cos, floor, radians, sqrt, sin, degrees, atan
 from subprocess import Popen, PIPE
+import sys
 
 import numpy as np
 
 try:
     import alphashape
+    import shapely
 except ImportError:
     # only used for get_perimeter function
     pass
+
 
 from qcore.binary_version import get_unversioned_bin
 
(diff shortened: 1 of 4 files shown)


viktor76525 (approved):

qcore/srf.py
```diff
@@ -11,15 +11,19 @@
 
 from math import ceil, cos, floor, radians, sqrt, sin, degrees, atan
 from subprocess import Popen, PIPE
+import sys
 
 import numpy as np
 
 try:
     import alphashape
+    from alphashape import optimizealpha
```
jpa198, line 20: Why are we importing both the module and a specific function from the module? Should either be only the module or only the classes/functions we need

qcore/srf.py
```diff
@@ -11,15 +11,19 @@
...
+    import shapely
```
jasonmotha, line 21: I think shapely is common enough that it doesn't need to be treated separately.

jasonmotha (changes requested):
Happy to have people disagree. But in general I don't like special cases of imports.

qcore/srf.py
sungeunbae: Agreed -

qcore/srf.py
sungeunbae: Agreed

viktor76525 (approved):

jasonmotha (approved):

jpa198 (approved):

PR #256, opened 2021-10-19
Instead of trying to load the whole VM file, we now check individual …

jpa198:
…layers one at a time. This should resolve core dump issues that have been plaguing automated VM gen

### DIFF
--- a/qcore/validate_vm.py
+++ b/qcore/validate_vm.py
@@ -18,14 +18,13 @@ import argparse
 from pathlib import Path
 
 import matplotlib.path as mpltPath
+import numpy as np
+
 from qcore.utils import load_yaml
 from qcore.constants import VM_PARAMS_FILE_NAME, VMParams
 from qcore.srf import get_bounds
 from qcore.geo import ll_dist, compute_intermediate_latitudes, build_corners
-
-import numpy as np
-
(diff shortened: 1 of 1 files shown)


qcore/validate_vm.py
```diff
@@ -213,26 +216,26 @@ def validate_vm_bounds(polygon, srf_bounds=None):
...
         )
-    if (
-        not np.min(
-            np.fromfile(file_name, dtype="<f{}".format(SIZE_FLOAT), count=vm_size)
-        )
-        > 0
-    ):
-        errors.append(f"File {file_name} has minimum value of 0.0")
+
+    with VelocityModelFile(nx, ny, nz, file_name, memmap=True) as vmf:
```
shot0829, line 234: just to be safe, probably add a
writable=False
into this?

shot0829 (changes requested):

qcore/validate_vm.py
shot0829: its the default behavior currently, but since we are sure we only need it to be read-only, adding the flag wont hurt.

shot0829 (approved):

jasonmotha (approved):

PR #259, opened 2021-11-15
Add single frequency bandpass to bwfilter

jpa198:


qcore/timeseries.py
```diff
@@ -57,6 +57,8 @@ def bwfilter(data, dt, freq, band, match_powersb=True):
     if match_powersb:
         if band == "highpass":
             freq *= 0.8956803352330285
+        elif band == "bandpass":
+            freq = np.asarray((freq*0.8956803352330285, freq*1.1164697500474103))
```
sungeunbae, line 61: Should we make them variables? I find it slightly uncomfortable to see repeating hard-coded constants.

sungeunbae (commented):
It's not a request change - just an opinion

qcore/timeseries.py
jasonmotha: I think we should. Especially considering we're using the same number in two places.

jpa198: Generally I agree with that, but this is simple enough that I'm happy to leave this as technical debt that potentially never gets resolved

jasonmotha: And additionally this is functionality that probably won't be used very often. :upside_down_face:

qcore/timeseries.py
```diff
@@ -54,13 +56,15 @@ def bwfilter(data, dt, freq, band, match_powersb=True):
     #    x *= -1
     # freq *= exp(x * math.log(sqrt(2.0) - 1.0))
     nyq = 1.0 / (2.0 * dt)
+    highpass_shift = 0.8956803352330285
+    lowpass_shift = 1.1164697500474103
```
jasonmotha, line 60: Just to be annoying, aren't these better as module level constants?

jpa198: They're used in one function, so I don't think so

jasonmotha (approved):

sungeunbae (approved):

PR #263, opened 2022-01-27
changed the formats file to read comma separated z file

jasonmotha:
I've checked this file and it looks like that the only place that this function was called was empirical db in seistech.

The proposed format going forward is keeping the comma delimited file (outputted from the VM code). - this is different to the tab/space delimited station and vs30 file.

Thoughts / changes?

### DIFF
--- a/qcore/formats.py
+++ b/qcore/formats.py
@@ -101,7 +101,7 @@ def load_z_file(z_file: str):
     :return: pd.DataFrame
         station as index and columns z1p0, z2p5
     """
-    return pd.read_csv(z_file, sep="\s+", index_col=0, header=None, names=["z1p0", "z2p5"])
+    return pd.read_csv(z_file, names=["z1p0", "z2p5", "sigma"], index_col=0, skiprows=1)
 
 
 def load_station_ll_vs30(station_file: str, vs30_file: str):


jpa198:
No issue from me with changing it, I'll leave it to the gm hazard group

SnowNooDLe (approved):

SnowNooDLe:
Based on the file we saw this morning, non_uniform.z, this change does make sense

jasonmotha:
Oh, actually I just realised changing this way will mean that Joel's stuff won't work. :thinking:

claudio525 (approved):

claudio525:
@jasonmotha 
>Oh, actually I just realised changing this way will mean that Joel's stuff won't work. thinking
We should decide on one format (comma separated imo), and any files not of that format should just be converted (should be pretty straightforward?)

jasonmotha:
> @jasonmotha
> 
> > Oh, actually I just realised changing this way will mean that Joel's stuff won't work. thinking
> 
> We should decide on one format (comma separated imo), and any files not of that format should just be converted (should be pretty straightforward?)

Yeah, agreed. Do you mean one format for all station files? Or one format for Z specifically?

Historically, station files were tab delimited and workflow depends on them being so. But Z is currently only used in gm_hazard - with potential to be used in standard empirical calculations. 
I think that's why the formats Z reader was tab delimited. 

Either way, I think chatting about this on Monday would probably be best.

claudio525:
@jasonmotha We forgot to chat regarding the above today. 
I recon ideally we would to comma separated across the board, but since thats not really feasible (due to the dependencies), I think just do comma separated for Z-files now. And then if the opportunity ever comes along to change the other files to comma separated as well, we should do it then.

jasonmotha:
@claudio525 Might be a good idea to chat about it tomorrow. Basically, I doubt we'd ever get to the point of updating station/vs30 files as they are used in EMOD3D.

Just had a quick look at project.py and I'll have to do a quick change there too.

joelridden (approved):

PR #264, opened 2022-02-01
Site specific bb

jpa198:


jasonmotha:
In general, looks good to me. What's the best way to get a quick explanation of what's happening here?

jpa198:
> In general, looks good to me. What's the best way to get a quick explanation of what's happening here?

Not even up for review yet, just wanted to get the diff...

The short of it is: The correct transfer function was included in Felipes source code for running site specific, and as part of that I chose to fix the function here (previous had no references) instead of adding it into workflow. 
Not sure if it'll be useful outside site specific, but we can refactor later if it becomes an issue for some reason.

Added the site specific entry to the platform config so we can have an organised way of storing the data

jpa198:
> In general, looks good to me. What's the best way to get a quick explanation of what's happening here?

These are the changes I've made to qcore to support the inclusion of site specific amplification through Chris/Felipes OpenSees work.
The transfer function details are as taken from the code Felipe provided for integration, and instead of add it somewhere else, I fixed the current function instead.
As to how it works, that's a question for Chris/Felipe, however I have used this as the transfer function for site specific runs that have been verified by Felipe, so I believe it is correct.

jasonmotha (approved):

sungeunbae (approved):

jasonmotha (approved):

joelridden (approved):

PR #265, opened 2022-03-09
Get distances multiple references

joelridden:
Allows for multiple reference lat and lon values to be used in the get distances calculation.
Ensured output was the same shape for using single float values too and works with numba.njit

### DIFF
--- a/qcore/geo.py
+++ b/qcore/geo.py
@@ -17,34 +17,35 @@ class InputError(Exception):
     pass
 
 
-def get_distances(locations: np.ndarray, lon: float, lat: float):
-    """Calculates the distance between the array of locations and
-    the specified reference location
+def get_distances(locations: np.ndarray, lon: np.float, lat: np.float):
+    """
+    Calculates the distance between the array of locations and
+    the specified reference location / locations
 
     Parameters
     ----------
     locations : np.ndarray
         List of locations
         Shape [n_locations, 2], column format (lon, lat)
-    lon : float
-        Longitude of the reference location
-    lat
-        Latitude of the reference location
+    lon : np.float
+        Array or singular float of Longitude reference locations to compare
+    lat : np.float
+        Array or singular float of Latitude reference locations to compare
 
     Returns
     -------
     np.ndarray
-        The distances, shape [n_locations]
+        The distances, shape [n_references, n_locations]
     """
     d = (
-        np.sin(np.radians(locations[:, 1] - lat) / 2.0) ** 2
-        + np.cos(np.radians(lat))
-        * np.cos(np.radians(locations[:, 1]))
-        * np.sin(np.radians(locations[:, 0] - lon) / 2.0) ** 2
+            np.sin(np.radians(np.expand_dims(locations[:, 1], axis=1) - lat) / 2.0) ** 2
+            + np.cos(np.radians(lat))
+            * np.cos(np.radians(np.expand_dims(locations[:, 1], axis=1)))
+            * np.sin(np.radians(np.expand_dims(locations[:, 0], axis=1) - lon) / 2.0) ** 2
     )
     d = R_EARTH * 2.0 * np.arctan2(np.sqrt(d), np.sqrt(1 - d))
 
-    return d
+    return d.T
 
 
 def closest_location(locations, lon, lat):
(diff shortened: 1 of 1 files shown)


jasonmotha (commented):
Looks good. Does it still work if we pass in a standard float to the function? Or do we need some input sanitation to maintain compatibility with existing code?

joelridden:
> Looks good. Does it still work if we pass in a standard float to the function? Or do we need some input sanitation to maintain compatibility with existing code?

It now is

jasonmotha (dismissed):

qcore/geo.py
```diff
@@ -18,33 +19,59 @@ class InputError(Exception):
...
     Returns
     -------
     np.ndarray
         The distances, shape [n_locations]
     """
+    return get_multiple_distances(locations, lon, lat)[0]
+
+
+@njit
+def get_multiple_distances(locations: np.ndarray, lon: np.float, lat: np.float):
```
claudio525, line 47: Whats the difference between the two functions? Just combine them into one?
Also does @njit actually do much in this case since its vectorized?

claudio525 (approved):

claudio525 (commented):
The previous one was meant to be a comment not approve..

qcore/geo.py
jasonmotha: One returns a vector of vectors. One returns a vector. Just so get_distances returns the same datastructure for the same inputs (for compatibility reasons)

I'll let Joel answer the performance question - but I'd expect that since the original had a decent improvement with njit that it still would have that improvement.

jpa198 (dismissed):

qcore/geo.py
```diff
@@ -7,6 +7,7 @@
 from warnings import warn
 
 import numpy as np
+from numba import njit
```
jpa198, line 10: Need to add numba to the requirements file as it's a new requirement for the repo

qcore/geo.py
joelridden: So turns out that the njit doesn't do much in terms of performance tested with multiple faults averaging 20 times.
image

This could be removed, except then would have to go through GMHazard, IM_calc to remove the numba.njit calls to the get_distances function. It is easier to leave it, but willing to do either.

Thoughts @claudio525 ?

jasonmotha (dismissed):

qcore/geo.py
claudio525: Now that their is no numba dependence, thoughts on combining these two functions? @jasonmotha @joelridden @jpa198

jasonmotha: I mean you can do one function called get_distances that has a split return type depending if sites is a vector or not.

I think what we have at the moment is the most elegant soln. though.

joelridden: I think they should combine and have split return types. get_distances feels like its taking alot of undeeded lines up

claudio525: How come this was resolved? 
I def think that they should be combined, two functions for the same thing seems overkill. And its not like this is uncommon, e.g. numpy functions return a float if you pass floats in, but if you pass arrays in then you get arrays back

qcore/geo.py
```diff
@@ -68,7 +46,7 @@ def get_multiple_distances(locations: np.ndarray, lon: np.float, lat: np.float):
         * np.sin(np.radians(np.expand_dims(locations[:, 0], axis=1) - lon) / 2.0) ** 2
     )
     d = R_EARTH * 2.0 * np.arctan2(np.sqrt(d), np.sqrt(1 - d))
-    return d.T
+    return d.T[0] if d.shape[1] == 1 else d.T
```
jasonmotha, line 49: Can we move the transpose to line 48?

jasonmotha (approved):

jasonmotha (approved):

qcore/geo.py
```diff
@@ -17,35 +17,12 @@ class InputError(Exception):
     pass
 
 
-def get_distances(locations: np.ndarray, lon: float, lat: float):
+def get_distances(locations: np.ndarray, lon: np.float, lat: np.float):
```
claudio525, line 20: Docstring needs to be updated slightly since the function now supports multiple reference locations
And the type hints for lon & lat should be `Union[float, np.ndarray]`
Also I think in general its better to use float than np.float

jpa198 (dismissed):

jasonmotha (approved):

claudio525 (approved):

jpa198 (approved):

PR #266, opened 2022-03-16
Transfer modules

SnowNooDLe:
# Transfer modules

Related to this [PR](https://github.com/ucgmsim/Pre-processing/pull/199), import some modules from pre-processing to achieve the following workflow:

image


jasonmotha:
nhm.py in qcore should have some import changes too?

jasonmotha (approved):

SnowNooDLe:
> nhm.py in qcore should have some import changes too?

Yup, it does

However, I kept the way qcore does instead of what we do in GMHazard regarding import modules
I was thinking to change to the import module but I do not want to interrupt the conventions we have within qcore :+1: 

image

jpa198 (approved):

claudio525 (approved):

PR #267, opened 2022-03-29
Adding WCT as a Status Enum

joelridden:
As part of the workflow metadata refactor, the WCT needed to be added to the Status Enum

### DIFF
--- a/qcore/constants.py
+++ b/qcore/constants.py
@@ -274,7 +274,7 @@ class Status(ExtendedStrEnum):
     running = 3, "running"
     unknown = 4, "unknown"
     completed = 5, "completed"
-    WCT = 6, "WCT"
+    killed_WCT = 6, "killed_WCT"
     failed = 7, "failed"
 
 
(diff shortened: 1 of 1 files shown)


jasonmotha (dismissed):
Pending https://github.com/ucgmsim/slurm_gm_workflow/pull/433#event-6322770019 being merged at the same time

sungeunbae (commented):
WCT just refers to "wall clock time", not directly indicating the status. what about timesup or an adjective carrying similar meaning, rather than a noun?

jpa198 (dismissed):

qcore/constants.py
```diff
@@ -274,7 +274,8 @@ class Status(ExtendedStrEnum):
     running = 3, "running"
     unknown = 4, "unknown"
     completed = 5, "completed"
-    failed = 6, "failed"
+    killed_WCT = 6, "killed_WCT"
```
jpa198, line 277: Excellent, a good choice of prefix

jasonmotha (approved):
Good call to add the suffix after we decided on slaughtered / killed.

sungeunbae (approved):

PR #269, opened 2022-07-25
get_fault_from_realisation() fix for a fault name containing an _

sungeunbae:
If a fault name contains an underscore, get_fault_from_realisaton can split the fault name and realisation part incorrectly.
For example, get_fault_from_realisaton_fix(XXXX_YYY_REL01) should return XXXX_YYY, not XXXX.


sungeunbae:
Merging into the forked repo first.

PR #270, opened 2022-07-27
get_fault_from_realisation() fix for a fault name containing an underscore

sungeunbae:
If a fault name contains an underscore, get_fault_from_realisaton can split the fault name and realisation part incorrectly.
For example, get_fault_from_realisaton_fix(XXXX_YYY_REL01) should return XXXX_YYY, not XXXX.

### DIFF
--- a/qcore/simulation_structure.py
+++ b/qcore/simulation_structure.py
@@ -10,7 +10,6 @@ def get_fault_from_realisation(realisation):
     realisation = os.path.basename(realisation)  # if realisation is a fullpath
     return realisation.rsplit("_REL",1)[0]
 
-
 def get_realisation_name(fault_name, rel_no):
     return "{}_REL{:0>2}".format(fault_name, rel_no)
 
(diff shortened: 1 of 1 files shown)


jpa198 (approved):

qcore/simulation_structure.py
```diff
@@ -8,7 +8,7 @@
 
 def get_fault_from_realisation(realisation):
     realisation = os.path.basename(realisation)  # if realisation is a fullpath
-    return realisation.split("_")[0]
+    return realisation.split("_REL")[0]
```
jasonmotha, line 11: Can we change this to rsplit("_REL", 1)[0]?

```
In [6]: test                                                                                        
Out[6]: 'MT_RELERSON_REL01'

In [7]: test.rsplit("_REL",1)[0]                                                                    
Out[7]: 'MT_RELERSON'
```

I'm not 100% convinced that other parts of the workflow can handle underscores in the fault name but we can cross that bridge later.

jasonmotha:
The `, 1` in the rsplit ensures it on the first one it finds.

jasonmotha (approved):

jpa198 (approved):

PR #272, opened 2022-08-22
Support for old non_REL realisation names

sungeunbae:
Related to an old PR https://github.com/ucgmsim/qcore/pull/270

An old realisation name not including "_REL" can not correctly return a fault name and this fix resolves it.
eg. Hossack_HYP01-501_S1244 -> Hossack


jpa198:
Do we need support for this other than e2e test data? (Which should probably be updated to use current standards instead of grandfathering in functionality)
Old simulation runs that we want to look at should either use an old copy of workflow, or we should redo it with the current naming convention

sungeunbae:
True - it's E2E tests I have a trouble with. I agree that an update to E2E tests is a better direction I should take. Happy to close this PR if we add E2E test update to the next sprint.

jasonmotha:
Agreed. However, does anyone have capacity to do that next sprint?

jpa198:
This should be closed in favour of updating the E2E test data

PR #278, opened 2022-11-28
Median event generation

jpa198:


qcore/constants.py
```diff
@@ -105,6 +105,58 @@ def iterate_str_values(cls, ignore_none=True):
...
+        elif isinstance(self._process, tuple):
+            self._process = ProcessType(self._process[0])
+        return self._process
+
+    @property
+    def dependency(self):
+        if self._dependency is True:
+            self._dependency = DependencyTarget.MEDIAN
+        elif self._dependency is False:
+            self._dependency = DependencyTarget.REL
```
joelridden, line 143: Probably don't need the extra Boolean for dependency, can just be a default to DependencyTarget.REL

qcore/constants.py
```diff
@@ -197,36 +330,31 @@ def __new__(
...
+        if any(
+            (
+                all((x in completed_dependencies for x in multi_dependency))
+                for multi_dependency in self.dependencies
+            )
+        ):
+            # At least one of the dependency conditions for the task is fulfilled, no need to add any more tasks
+            return []
+        # Otherwise the first dependency list is the default
+        return [x for x in self.dependencies[0] if x not in completed_dependencies]
```
joelridden, line 350: Way nicer

qcore/constants.py
```diff
@@ -151,38 +232,86 @@ class ProcessType(ExtendedStrEnum):
...
         False,
         "time python $IMPATH/calculate_ims.py {sim_dir}/BB/Acc/BB.bin b -o {sim_dir}/IM_calc/ -np {np} -i "
         "{sim_name} -r {fault_name} -t s {component} {extended} {simple} {advanced_IM} {pSA_periods}",
-        ((5,), (12,), (13,)),
+        (
+            (Dependency(BB, DependencyTarget.REL),),
+            (Dependency(LF2BB, DependencyTarget.REL),),
+            (Dependency(HF2BB, DependencyTarget.REL),),
+        ),
     )
```
sungeunbae, line 263: I have a bit of problem with the readability.
```
    BB = (
        5,
...
        (
            (
                Dependency(EMOD3D, DependencyTarget.REL),
                Dependency(HF, DependencyTarget.REL),
            ),
        ),
    )
```
and
```
    IM_calculation = (
        6,
        "IM_calc",
...
        (
            (Dependency(BB, DependencyTarget.REL),),
            (Dependency(LF2BB, DependencyTarget.REL),),
            (Dependency(HF2BB, DependencyTarget.REL),),
        ),
    )
```
For BB, both dependencies should be met, and IM_calculation, either of 3 can be met. 
Here, the multiple layers of tuples makes things quite confusing and less intuitive. Also, I find it a bit too verbose.

Can we make it something like

```
    BB = (
        5,
...
        (
            [EMOD3D.REL, HF.REL]
        ),
    )
```
and

```
    IM_calculation = (
        6,
     ...
        (
            [BB.REL],
            [LF2BB.REL],
            [HF2BB.REL],
        ),
    )
```
where [] used for all and () used for any. (Either way is fine) 
Regarding the dot in LF2BB.REL, it's just an example of possible simplification.  I am sure there is a better alternative notation.

joelridden (approved):

sungeunbae (approved):

PR #281, opened 2023-01-13
Automated srf gen

jpa198:


qcore/constants.py
```diff
@@ -164,7 +164,33 @@ class ProcessType(ExtendedStrEnum):
     # ProcessID, ProcessName, "Does this task use Hyperthreading?", "Does this use an Acc directory?", command, dependancies (tuple),
     """
 
-    VM_PARAMS = 16, "VM_PARAMS", None, False, None, ([],)
+    # GCMT_2_REL = 24, "GCMT_2_REL", None, False, None, ([],)
+    # NHM_2_REL = 23, "NHM_2_REL", None, False, None, ([],)
```
joelridden, line 168: Do the commented out lines need to be here?

qcore/constants.py
```diff
@@ -179,7 +205,15 @@ class ProcessType(ExtendedStrEnum):
         None,
         False,
         None,
-        ([(VM_PARAMS, "MEDIAN")],),
+        ([(VM_PARAMS, "MEDIAN"), (INSTALL_REALISATION, "REL")],),
+    )  # Needs VM_params generated for REL_1
```
joelridden, line 209: Median not REL_01?

qcore/constants.py
```diff
@@ -179,7 +205,15 @@ class ProcessType(ExtendedStrEnum):
...
+    NO_VM_PERT = (
+        20,
+        "NO_VM_PERT",
+        None,
+        False,
+        None,
+        ([(INSTALL_REALISATION, "REL")],),
     )  # Needs VM_params generated for REL_1
```
joelridden, line 217: Doesn't need this comment anymore

qcore/constants.py
```diff
@@ -356,6 +393,8 @@ def check_mutually_exclusive_tasks(tasks: List["ProcessType"]):
         :return: A string containing any errors found during the check"""
         mutually_exclusive_tasks = (
             (ProcessType.BB, ProcessType.LF2BB, ProcessType.HF2BB),
+            # (ProcessType.GCMT_2_REL, ProcessType.NHM_2_REL),
```
joelridden, line 396: Is this comment needed

qcore/utils.py
```diff
@@ -123,20 +123,36 @@ def load_sim_params(sim_yaml_path, load_fault=True, load_root=True, load_vm=True
...
-    if load_root or load_vm and not load_fault:
+
+    if load_root is True or load_vm is True and not load_fault:
         load_fault = True  # root/vm_yamlpath in fault_yaml
-    if load_fault:
+
+    if sim_yaml_path:
+        sim_params = load_yaml(sim_yaml_path)
+    elif load_fault is True:
+        raise ValueError("For automated fault_params loading, sim_params must be set")
```
joelridden, line 137: Isn't sim_yaml_path always set as there is no default?

qcore/utils.py
```diff
@@ -123,20 +123,36 @@ def load_sim_params(sim_yaml_path, load_fault=True, load_root=True, load_vm=True
...
+
+    if load_fault is True:
         fault_params = load_yaml(sim_params["fault_yaml_path"])
-    if load_root:
+    elif load_fault:
+        fault_params = load_yaml(load_fault)
```
joelridden, line 142: Maybe now we should adjust the docstring a bit to know you can just pass in a path instead?

joelridden (commented):
Mainly just around comments

qcore/utils.py
jpa198: Good question.  sim_params.yaml doesn't exist before INSTALL_REL runs, so before then we have to load the individual params yamls somewhat manually.
Happy to take input on an improved method

qcore/utils.py
```diff
@@ -123,20 +123,36 @@ def load_sim_params(sim_yaml_path, load_fault=True, load_root=True, load_vm=True
...
     fault_params = {}
     root_params = {}
     vm_params = {}
-    if load_root or load_vm and not load_fault:
+
+    if load_root is True or load_vm is True and not load_fault:
         load_fault = True  # root/vm_yamlpath in fault_yaml
-    if load_fault:
+
+    if sim_yaml_path:
```
sungeunbae, line 134: Can we do all the new smart & direct load_yaml when this if test fails, instead of having them spread over?

sungeunbae (commented):
Perhaps will need to update the docstring, and change the order of the logic slightly to load all yamls if sim_params is not ready (ie. when load_fault, load_vm, load_root are not boolean but path)

qcore/constants.py
jpa198: They're there as pre-fill for automated realisation generation, to reduce the work later on.
Don't have to be there, but they show the way forward and I don't think it hurts to have them there

qcore/constants.py
jpa198: Related to automated realisation generation, see above comments

qcore/utils.py
jpa198: Do you mean, have the old style loading for when everything is available, and have a separate/new style method of loading for when there are only partial params files available?
I think this is neater/more concise than having two methods, but I'm open to input if others are onboard with it

joelridden (approved):

qcore/utils.py
sungeunbae: I meant, something like this. 
```
if sim_yaml_path: # which is when load_fault, load_root, load_vm are boolean
    sim_params = load_yaml(sim_yaml_path)
    if load_fault:
        fault_params = load_yaml(sim_params["fault_yaml_path"])
    if load_root:                
       root_params = load_yaml(fault_params["root_yaml_path"])
    if load_vm:       
       vm_params = load_yaml(
            os.path.join(fault_params["vel_mod_dir"], "vm_params.yaml")
        )
 else: # when load_fault, load_root, load_vm are paths
    if load_fault:
        fault_params = load_yaml(load_fault)
    if load_root:        
       root_params = load_yaml(load_root)
    if load_vm:
       vm_params = load_yaml(load_vm)
```

sungeunbae (approved):

PR #283, opened 2023-03-15
Archive Structure

joelridden:
Adding functions to support new archive structure

### DIFF
--- a/qcore/archive_structure.py
+++ b/qcore/archive_structure.py
@@ -5,10 +5,15 @@ from pathlib import Path
 
 
 def get_fault_from_realisation(realisation: str):
-    """ Gets the fault name from the realisation name"""
+    """Gets the fault name from the realisation name"""
     return realisation.rsplit("_REL", 1)[0]
 
 
 def get_IM_csv_from_root(archive_root: Path, realisation: str):
-    """ Gets the full path to the im_csv file given the archive root dir and the realistion name"""
-    return archive_root / get_fault_from_realisation(realisation) / "IM" / f"{realisation}.csv"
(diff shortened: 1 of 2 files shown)


qcore/archive_structure.py
```diff
@@ -0,0 +1,20 @@
+"""
+Gives access to the folder structure of archived cybershake directories
+"""
+from pathlib import Path
+
+
+def get_fault_from_realisation(realisation: str):
+    """Gets the fault name from the realisation name"""
+    return realisation.rsplit("_REL", 1)[0]
```
jpa198, line 9: If this function is already in simulation structure, it should be imported instead of re-written. To make sure they do the same thing

jpa198 (dismissed):

sungeunbae (approved):

jpa198 (dismissed):

jpa198 (approved):

qcore/archive_structure.py
```diff
@@ -0,0 +1,17 @@
...
+from simulation_structure import get_fault_from_realisation
+
+
+def get_IM_csv_from_root(archive_root: Path, realisation: str):
+    """Gets the full path to the im_csv file given the archive root dir and the realistion name"""
+    fault_name = get_fault_from_realisation(realisation)
+    return (
+        archive_root
+        / fault_name
+        / f"{fault_name}_IM"
```
sungeunbae: We agreed to have the fault_name in .tar, file, eg)  {fault_name}_IM.tar, but when untarred, and recreate the archive directory hierachy, it will be archive_root/ fault_name/ IM, without the  {fault_name} prefix from the IM subdirectory name.

sungeunbae (approved):

PR #296, opened 2023-12-19
Added a support for loading station file of random format

sungeunbae:
Occasionally we get a station file of random format (unlike .ll, .vs30 files) to feed into other in-house developed scripts. 
For example, someone gave me a CSV file and I wanted to load this file, make a dataframe and do further processing, and we have no support from formats.py. 

```
Station Code,Latitude,Longitude,Basin_NZVM 2.07
ADCS,-43.9024444,171.747563,NewCanterburyBasinBoundary_WGS84_1m
AICS,-42.76818509,171.6243571,
AKSS,-43.81091,172.96349,BPVBoundary
AMBC,-43.15467772,172.7308795,NewCanterburyBasinBoundary_WGS84_1m
APPS,-42.94894944,171.5677355,
APZ,-46.83180552,167.9888255,
...
```
Of course, we can use pd.read_csv() and utilize its rich customisation options, but when you get to process yet another file with a different format, you need to update your code, how you call pd.read_csv().  I wanted a way we can specify the format at run time without needing to edit the code.

I added a new load_generic_station_file() function, which works like a standard load_station_file() if no arguments are configured.

Also supplied an assisting function station_file_argparser(), that faciliates obtaining format-related info from the user at the time of a script execution.

In your script, X.py, you already have some arguments parsed by ArgumentParser(), but if you wish to handle extra arguments related to the format of a station file
```    

    def get_args():
        parser = argparse.ArgumentParser()
        arg = parser.add_argument
        arg("--arg1", help="argument 1")
        arg("--arg2", help="argument 2")
        parser = formats.station_file_argparser(parser=parser)  # pass the parser argument, update it with the return value
        return parser.parse_args()

    if __name__ == '__main__':
        args = get_args()
        ....
```
and you can run your script like this

```
    python X.py --arg1 ARG1 --arg2 ARG2 --stat_file some_station_file.csv --stat_name_col 0 --lat_col 1 --lon_col 2 --sep , --skiprows 1
```
```

### DIFF
--- a/qcore/formats.py
+++ b/qcore/formats.py
@@ -180,8 +180,10 @@ def load_generic_station_file(
 
     return pd.read_csv(
         stat_file,
-        usecols=cols.values(),
-        names=sorted(cols, key=cols.get),
+        usecols=cols.values(),  # we will be loading columns of these indices (order doesn't matter)
+        names=sorted(
+            cols, key=cols.get
+        ),  # eg. cols={stat_name:2, lon:0, lat:1} means names = ["lon","lat","stat_name"]
         index_col=stat_name_col,
         sep=sep,
         header=None,
--- a/qcore/formats.py
+++ b/qcore/formats.py
@@ -4,6 +4,7 @@ Functions and classes to load data that doesn't belong elsewhere.
 
 import pandas as pd
 import numpy as np
+import argparse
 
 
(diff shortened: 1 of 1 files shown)


joelridden (approved):
I like this idea, can see how it could be useful

claudio525 (approved):

PR #297, opened 2024-03-27
Point in Polygon

joelridden:
Adding functions for finding out if a point or set of points are in a polygon.
Only thought is the return 2 about being on an edge I'm not sure about, especially if we are using this as a true and false function

### DIFF
--- a/qcore/point_in_polygon.py
+++ b/qcore/point_in_polygon.py
@@ -1,0 +1,75 @@
+from numba import jit, njit
+import numba
+import numpy as np
+
+
+@jit(nopython=True)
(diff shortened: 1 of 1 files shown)


claudio525 (commented):
Add the link to the stackoverflow question where this solution came from, otherwise this looks great.

sungeunbae:
> Add the link to the stackoverflow question where this solution came from, otherwise this looks great.

https://github.com/sasamil/PointInPolygon_Py/blob/master/pointInside.py

sungeunbae (approved):

claudio525 (approved):

PR #300, opened 2024-04-16
Add functions to compute the closest points between planes in 3D

lispandfound:
Adds two key functions `closest_points_between_planes` and `closest_points_between_plane_sequences` that compute the two closest points between two planes and two connected sequences of planes, respectively. This is necessary to enable a highly parallelised computation of type-5 SRF files. The main function, `closest_points_between_planes` uses the constrained optimisers from scipy, and relies on some geometry to compute the constraints required to keep the two points on the planes. Where such geometry may not be known to the reader, I have added some explanatory comments.

NB: because this code uses an optimiser rather than explicitly solving for the closest points, the computed points are only approximate. My tests seem to indicate that each coordinate is no more than 1e-7 away from the correct coordinates in absolute value (see the test file).

### DIFF
--- a/qcore/geo.py
+++ b/qcore/geo.py
@@ -901,10 +901,12 @@ def projective_span(p: np.ndarray, q: np.ndarray, r: np.ndarray) -> np.ndarray:
     return ns / c
 
 
-def plane_from_three_points(p: np.ndarray, q: np.ndarray, r: np.ndarray):
+def plane_from_three_points(p: np.ndarray, q: np.ndarray, r: np.ndarray) -> np.ndarray:
     """Find the plane spanning three points.
 
     Returns the coefficient vector of the affine plane spanning three points.
+    Note this is distinct from projective_span in that p, q, and r are three
+    points in R^3.
 
     Parameters
     ----------
@@ -914,6 +916,11 @@ def plane_from_three_points(p: np.ndarray, q: np.ndarray, r: np.ndarray):
         a point on the plane.
     r : np.ndarray
         a point on the plane.
+
+    Returns
+    -------
+    np.ndarray
+        The dual coordinates of the plane spanning p, q, and r.
 
     Examples
     --------
--- a/qcore/geo.py
+++ b/qcore/geo.py
@@ -854,7 +854,7 @@ def homogenise_point(p: np.ndarray) -> np.ndarray:
     array([1,0,0,1])
     """
 
-    return np.concatenate((p, np.array([1])))
+    return np.append(p, 1)
 
 
 def projective_span(p: np.ndarray, q: np.ndarray, r: np.ndarray) -> np.ndarray:
--- a/qcore/geo.py
+++ b/qcore/geo.py
@@ -904,7 +904,7 @@ def projective_span(p: np.ndarray, q: np.ndarray, r: np.ndarray) -> np.ndarray:
     #
     # NOTE: This choice to normalise by the last coordinate is arbitrary. Any
     # method of normalisation would work.
-    null_space = null_space.reshape((1, -1))[0]
+    null_space = null_space.ravel()
     c = next(x for x in reversed(null_space) if not np.isclose(x, 0))
     return null_space / c
 
--- a/qcore/geo.py
+++ b/qcore/geo.py
@@ -8,6 +8,8 @@ from warnings import warn
 from typing import Union
 
 import numpy as np
+import scipy as sp
+import functools
 
 from qcore.binary_version import get_unversioned_bin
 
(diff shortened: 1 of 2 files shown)


qcore/geo.py
```diff
@@ -827,3 +830,300 @@ def compute_intermediate_latitudes(lon_lat1, lon_lat2, lon_in):
...
+    np.ndarray
+        The associated point in PG(3, R).
+
+    Examples
+    --------
+    >>> homogenise_point(np.array([1, 0, 0]))
+    array([1,0,0,1])
+    """
+
+    return np.concatenate((p, np.array([1])))
```
joelridden, line 857: could this just be np.append(p, 1)?

qcore/geo.py
```diff
@@ -827,3 +830,300 @@ def compute_intermediate_latitudes(lon_lat1, lon_lat2, lon_in):
...
+    # not span a unique plane. This could happen if the points supplied are
+    # collinear.
+    if ns.shape[1] != 1:
+        raise ValueError("Points supplied do not span a plane.")
+    # Otherwise we just rescale the homogenous vector. It is now mathematically
+    # impossible that the vector is zero, so we can safely assume that c != 0 in the following code.
+    # We normalise by the last non-zero coordinate of the point.
+    ns = ns.reshape((1, -1))[0]
+    c = next(x for x in reversed(ns) if not np.isclose(x, 0))
+    return ns / c
```
joelridden, line 901: Maybe some easier to understand variable names other than p, q, r, M, ns, c and x could be used? Unless there is a mathematical definition reason for doing so?

qcore/geo.py
```diff
@@ -827,3 +830,300 @@ def compute_intermediate_latitudes(lon_lat1, lon_lat2, lon_in):
...
+
+
+def plane_from_three_points(p: np.ndarray, q: np.ndarray, r: np.ndarray):
```
sungeunbae, line 904: return type hint?

qcore/geo.py
```diff
@@ -827,3 +830,300 @@ def compute_intermediate_latitudes(lon_lat1, lon_lat2, lon_in):
...
+            p1_ortho_planes[i] *= -1
+
+    # We now repeat this process for p2.
+    p2_centroid = np.average(p2_corners, axis=0)
+    p2_ortho_planes = [
+        orthogonal_plane(
+            p2_dual_coordinates, p2_corners[i], p2_corners[(i + 1) % len(p2_corners)]
+        )
+        for i in range(len(p2_corners))
+    ]
```
joelridden, line 1053: Since this is a repeated process could this and the explanation be it's own function and just done for both p1 and p2

qcore/geo.py
```diff
@@ -827,3 +830,300 @@ def compute_intermediate_latitudes(lon_lat1, lon_lat2, lon_in):
...
+            "fun": functools.partial(p1_constraint, ortho_coords),
+        }  # we use functools.partial because we want to copy ortho_coords by value rather than by reference.
+        for ortho_coords in p1_ortho_planes
+    ]
+    p2_bounds_constraints = [
+        {"type": "ineq", "fun": functools.partial(p2_constraint, ortho_coords)}
+        for ortho_coords in p2_ortho_planes
+    ]
+
+    def distance(x):
```
joelridden, line 1076: position_of_points maybe instead of x?

qcore/geo.py
```diff
@@ -827,3 +830,300 @@ def compute_intermediate_latitudes(lon_lat1, lon_lat2, lon_in):
...
+        # x is a numpy array of length 6 containing the position of the points p in x[:3] and q in x[3:].
+        # returns ||p - q||^2
+        return sp.spatial.distance.cdist(
+            x[:3].reshape((1, 3)), x[3:].reshape((1, 3)), metric="sqeuclidean"
+        )
+
+    # Our initial guess is just the centroid of each plane.
+    x0 = [*p1_centroid, *p2_centroid]
```
joelridden, line 1084: Maybe a better variable name for x0? like initial_guess?

qcore/test/test_geo/test_geo.py
```diff
@@ -106,3 +107,135 @@ def test_wgs_nztm2000x(test_lonlat, output_points):
     test_points = geo.wgs_nztm2000x(test_lonlat)
     sample_output_points = np.array(output_points)
     utils.compare_np_array(test_points, sample_output_points)
+
```
joelridden, line 110: This is great!

joelridden (commented):
Mainly just variable name things I would say for easier understanding and maybe another function, but great explanations!

qcore/geo.py
sungeunbae: I found this part difficult to follow.  ns is supposed to have dimension 1 to span a unique plane, and we check ns.shape[1] != 1..? Could you also explain why "dividing by the last non-zero coordinate" makes it normalized?

lispandfound: Right, it's better to think of the null space not as the span of p, q, and r, but as the space of all planes with dual coordinates x such that x * p == 0, x * q == 0, x * r == 0. In this context, if you have three distinct non-collinear points, they are contained within a unique plane with some dual coordinates x. However the dual coordinates for this plane are not unique. In particular, any scalar multiple of x will define the same plane.  Thus, the null space will have dimension 1 in this case (to represent all scalar multiples of x). 

The normalisation step is simply to pick one "good" representation for x. As discussed, while the plane spanning p, q, and r is unique, the dual coordinates for x are not necessarily unique. In particular, I have chosen to scale x such that the last non-zero coordinate is one (which is achieved by dividing through by that coordinate). This choice is arbitrary, and shares the same name as normalising by dividing through by the length of x it's entirely unrelated.

For an example of how this works, suppose we are trying to find the plane containing the points (0, 0, 0), (0, 1, 0) and (0, 0, 1). We first add the extra 1 coordinate:

(0, 0, 0, 1)
(0, 1, 0, 1)
(0, 0, 1, 1)

and then we compute the null space, which is one-dimensional and has a single basis vector (1, 0, 0, 0). This represents the plane with equation x = 0. Any scalar multiple of these coordinates would suffice as the dual coordinates, (2, 0, 0, 0) represents the equation 2x = 0 (which is the same plane). But because we need to make one choice and stick to it, the code will rescale the solution to get (1, 0, 0, 0).

I'll shorten this and change the description to make it clearer what is going on.

lispandfound: Oh yeah in the cases where the dimension of the null space is not 1, there are multiple inequivalent dual coordinates x that satisfy x * p, x * q, x * r == 0. e.g.

(1, 0, 0, 1)
(0, 1, 0, 1)
(1/2, 1/2, 0, 1)

These points are collinear, and the null space of that matrix has rank 2 and is spanned by the vectors (0, 0, 1, 0) and (1, 1, 0, -1). This is to represent the fact that there are multiple different planes that contain these points, e.g. y = 0 and 1/2 x + 1/2 y - 1 = 0.

sungeunbae (dismissed):

joelridden (approved):

sungeunbae (approved):

claudio525 (approved):

PR #303, opened 2024-05-12
Add gsf module

lispandfound:
@claudio525 Mentioned that the parsing code in ucgmsim/visualization#91 may have broader application and so should be included in qcore. I agree, so this pull request contains the GSF parsing code.

### DIFF
--- a/qcore/coordinates.py
+++ b/qcore/coordinates.py
@@ -29,12 +29,21 @@ def wgs_depth_to_nztm(wgs_depth_coordinates: np.ndarray) -> np.ndarray:
     Parameters
     ----------
     wgs_depth_coordinates : np.ndarray
-        An array of shape (N, 3) containing WGS84 coordinates (latitude, longitude, depth).
+        An array of shape (N, 3) or (3,) containing WGS84 coordinates
+        (latitude, longitude, depth).
 
     Returns
     -------
     np.ndarray
         An array of shape (N, 3) containing NZTM coordinates (x, y, depth).
+
+    Examples
+    --------
+    >>> wgs_depth_to_nztm(np.array([-43.5320, 172.6366, 0]))
+    array([5180040.61473068, 1570636.6812821 ,       0.        ])
+    >>> wgs_depth_to_nztm(np.array([[-36.8509, 174.7645, 100], [-41.2924, 174.7787, 100]]))
+    array([[5.92021456e+06, 1.75731133e+06, 1.00000000e+02],
+           [5.42725716e+06, 1.74893148e+06, 0.00000000e+00]])
     """
     if wgs_depth_coordinates.shape == (3,):
         return np.array(_WGS2NZTM.transform(*wgs_depth_coordinates))
@@ -54,12 +63,20 @@ def nztm_to_wgs_depth(nztm_coordinates: np.ndarray) -> np.ndarray:
     Parameters
     ----------
     nztm_coordinates : np.ndarray
-        An array of shape (N, 3) containing NZTM coordinates (x, y, depth).
+        An array of shape (N, 3) or (3,) containing NZTM coordinates (x, y, depth).
 
     Returns
     -------
     np.ndarray
         An array of shape (N, 3) containing WGS84 coordinates (latitude, longitude, depth).
+
+    Examples
+    --------
+    >>> nztm_to_wgs_depth(np.array([5180040.61473068, 1570636.6812821, 0]))
+    array([-43.5320, 172.6366, 0])
+    >>> wgs_depth_to_nztm(array([[5.92021456e+06, 1.75731133e+06, 100],
+                                 [5.42725716e+06, 1.74893148e+06, 0]]))
+    array([[-36.8509, 174.7645, 100], [-41.2924, 174.7787, 100]])
     """
     if nztm_coordinates.shape == (3,):
         return np.array(_NZTM2WGS.transform(*nztm_coordinates))
--- a/qcore/coordinates.py
+++ b/qcore/coordinates.py
@@ -36,10 +36,13 @@ def wgs_depth_to_nztm(wgs_depth_coordinates: np.ndarray) -> np.ndarray:
     np.ndarray
         An array of shape (N, 3) containing NZTM coordinates (x, y, depth).
     """
-    nztm_coords = np.array(
-        _WGS2NZTM.transform(wgs_depth_coordinates[:, 0], wgs_depth_coordinates[:, 1]),
+    return np.array(
+        _WGS2NZTM.transform(
+            wgs_depth_coordinates[:, 0],
+            wgs_depth_coordinates[:, 1],
+            wgs_depth_coordinates[:, 2],
+        )
     ).T
-    return np.append(nztm_coords, wgs_depth_coordinates[:, 2].reshape((-1, 1)), axis=-1)
 
 
 def nztm_to_wgs_depth(nztm_coordinates: np.ndarray) -> np.ndarray:
@@ -56,7 +59,10 @@ def nztm_to_wgs_depth(nztm_coordinates: np.ndarray) -> np.ndarray:
     np.ndarray
         An array of shape (N, 3) containing WGS84 coordinates (latitude, longitude, depth).
     """
-    wgs_coords = np.array(
-        _NZTM2WGS.transform(nztm_coordinates[:, 0], nztm_coordinates[:, 1]),
+    return np.array(
+        _WGS2NZTM.transform(
+            nztm_coordinates[:, 0],
+            nztm_coordinates[:, 1],
+            nztm_coordinates[:, 2],
+        )
     ).T
-    return np.append(wgs_coords, nztm_coordinates[:, 2].reshape((-1, 1)), axis=-1)
--- a/qcore/coordinates.py
+++ b/qcore/coordinates.py
@@ -1,0 +1,62 @@
+"""
+Module for coordinate conversions between WGS84 (latitude and longitude) and NZTM (New Zealand Transverse Mercator) coordinate systems.
+
+This module provides functions for converting coordinates between WGS84 and NZTM coordinate systems.
+See linz.govt.nz for a description of the NZTM coordinate system.
+
+Functions:
+- wgs_depth_to_nztm(wgs_depth_coordinates: np.ndarray) -> np.ndarray:
+    Converts WGS84 coordinates (latitude, longitude, depth) to NZTM coordinates.
+- nztm_to_wgs_depth(nztm_coordinates: np.ndarray) -> np.ndarray:
+    Converts NZTM coordinates (x, y, depth) to WGS84 coordinates.
+"""
+
+import numpy as np
+import pyproj
+
+# Module level conversion constants for internal use only.
+_WGS_CODE = 4326
+_NZTM_CODE = 2193
+
+_WGS2NZTM = pyproj.Transformer.from_crs(_WGS_CODE, _NZTM_CODE)
+_NZTM2WGS = pyproj.Transformer.from_crs(_NZTM_CODE, _WGS_CODE)
+
+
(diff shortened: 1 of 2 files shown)


claudio525 (approved):

lispandfound:
I have also added a new coordinate transformation module, and all the GSF related code from the type5 branch of Pre-processing. The coordinate transformation module is required for the GSF meshgrid generator. @claudio525 I know you approved the trivial parsing code, but it now includes quite a bit more. I will also add a description of the GSF file format to the wiki because I don't think it's documented anywhere.

qcore/gsf.py
```diff
@@ -0,0 +1,210 @@
...
+def write_fault_to_gsf_file(
+    gsf_file_handle: TextIO,
+    meshgrids: list[np.ndarray],
+    lengths: np.ndarray,
+    widths: np.ndarray,
+    strikes: np.ndarray,
+    dips: np.ndarray,
+    rakes: np.ndarray,
+    resolution: int,
+):
```
lispandfound, line 124: I don't like this function the way that it is currently. I have a `fault` module in type5 atm that encapsulates lengths, widths, strikes, dips, etc into a Fault object rather than this parallel array nonsense. Perhaps I should include this fault module into qcore? I think it has some merit as the geometry code I wrote there could be useful for other projects (like for example, the NSHMDB repo).

lispandfound: If I did that, I'd also do the fault module properly, with multiple ways of instantiating faults (from corners, from centroid + strike, dip, etc). This is because the NSHMDB can read fault corners from the database, but EMOD3D specifies faults with centroids, but fundamentally they represent the same object.

qcore/coordinates.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Returns
+    -------
+    np.ndarray
+        An array of shape (N, 3) containing NZTM coordinates (x, y, depth).
+    """
+    nztm_coords = np.array(
+        _WGS2NZTM.transform(wgs_depth_coordinates[:, 0], wgs_depth_coordinates[:, 1]),
+    ).T
+    return np.append(nztm_coords, wgs_depth_coordinates[:, 2].reshape((-1, 1)), axis=-1)
```
joelridden, line 42: This can be simplified to just
`nztm_coords = np.array(
        _WGS2NZTM.transform(wgs_depth_coordinates[:, 0], wgs_depth_coordinates[:, 1], wgs_depth_coordinates[:, 2])).T`

qcore/coordinates.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Returns
+    -------
+    np.ndarray
+        An array of shape (N, 3) containing WGS84 coordinates (latitude, longitude, depth).
+    """
+    wgs_coords = np.array(
+        _NZTM2WGS.transform(nztm_coordinates[:, 0], nztm_coordinates[:, 1]),
+    ).T
+    return np.append(wgs_coords, nztm_coordinates[:, 2].reshape((-1, 1)), axis=-1)
```
joelridden, line 62: Same thing as other function

qcore/gsf.py
joelridden: Not against having a shared Fault object.

We have in the past created a classdef file which contained this https://github.com/ucgmsim/Pre-processing/blob/f86c185c4a87ba13b3a413e78132e8ca22fa6fb3/VM/models/classdef.py#L4 (Used to also be one in empirical engine till it was removed for over complicating the problem)

Did not handle things very well but could be good to maybe bring these all together under 1 hood if implemented properly and isn't too much "overhead" just to create a fault object to get results from functions.

qcore/gsf.py
```diff
@@ -0,0 +1,210 @@
...
+        - strike
+        - dip
+        - rake
+        - slip (nearly always -1)
+        - init_time (nearly always -1)
+        - seg_no (the fault segment this point belongs to)
+    """
+    with open(gsf_filepath, mode="r", encoding="utf-8") as gsf_file_handle:
+        # we could use pd.read_csv with the skiprows argument, but it's not
+        # as versatile as simply skipping the first n rows with '#'
```
joelridden, line 188: I assume thats because n rows is an unknown length which is different per file?

qcore/gsf.py
```diff
@@ -0,0 +1,210 @@
...
+                f"{point[1]:11.5f} {point[0]:11.5f} {point[2] / 1000:11.5e} {strike_step:11.5e} {dip_step:11.5e} {strike:6.1f} {dip:6.1f} {rake:6.1f} {-1.0:8.2f} {-1.0:8.2f} {i:3d}\n"
+            )
+
+
+def read_gsf(gsf_filepath: str) -> pd.DataFrame:
+    """Parse a GSF file into a pandas DataFrame.
+
+    Parameters
+    ----------
+    gsf_file_handle : TextIO
```
joelridden, line 167: actually a str input?

sungeunbae (commented):
Happy in general - just a few comments

qcore/gsf.py
lispandfound: Yeah the number of comment lines prior to the number of points line is not defined. I generate GSF files with just one comment line, rob uses 3 lines, both are acceptable.

qcore/gsf.py
lispandfound: Ah yeah good catch

qcore/gsf.py
```diff
@@ -0,0 +1,210 @@
...
+    transformation_matrix = np.vstack(
+        [(x_upper - origin) / length_x, (y_bottom - origin) / length_y]
+    ).T
+    nztm_meshgrid = (transformation_matrix @ subdivision_coordinates).T
+    nztm_meshgrid += origin
+    return coordinates.nztm_to_wgs_depth(nztm_meshgrid)
+
+
+def write_fault_to_gsf_file(
+    gsf_file_handle: TextIO,
```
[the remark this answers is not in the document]
lispandfound: Yeah that should be a Path, you're right

lispandfound:
@sungeunbae @joelridden I have addressed these concerns. I'll leave the unhappy parallel array gsf writing function for now but include a fault class into qcore that I can use in NSHMDB, Pre-Processing and any places later on where the database or type-5 code gets incorporated.

sungeunbae (approved):

qcore/coordinates.py
joelridden: These comments would be nice to address?

joelridden (commented):
Just think the other 2 comments for the coordinates.py functions would be good to simplify

joelridden (approved):

qcore/gsf.py
```diff
@@ -0,0 +1,211 @@
...
+        Coordinates of the bottom y boundary (lat, lon, depth).
+    resolution : float
+        Resolution of the meshgrid.
+
+    Returns
+    -------
+    np.ndarray
+        The meshgrid of the rectangular planar region.
+    """
+    origin = coordinates.wgs_depth_to_nztm(origin)
```
claudio525, line 96: I think it might be worth breaking this up into some blocks, and then adding some block comments. There is a fair few steps and not all of them are trivial imo.

qcore/gsf.py
claudio525: This would line up with the `preprocessing` -> `source_model` repo change that we will hopefully do in the near future.

qcore/gsf.py
```diff
@@ -0,0 +1,211 @@
...
+        number_of_points = sum(meshgrid.shape[0] for meshgrid in meshgrids)
+        gsf_file_handle.write(f"{number_of_points}\n")
+        for i, (length, width, strike, dip, rake, meshgrid) in enumerate(
+            zip(lengths, widths, strikes, dips, rakes, meshgrids)
+        ):
+            strike_step = length / gridpoint_count_in_length(length * 1000, resolution)
+            dip_step = width / gridpoint_count_in_length(width * 1000, resolution)
+            for point in meshgrid:
+                gsf_file_handle.write(
+                    f"{point[1]:11.5f} {point[0]:11.5f} {point[2] / 1000:11.5e} {strike_step:11.5e} {dip_step:11.5e} {strike:6.1f} {dip:6.1f} {rake:6.1f} {-1.0:8.2f} {-1.0:8.2f} {i:3d}\n"
```
claudio525, line 159: This line is possibly a bit long? Black doesn't break up string lines, so has to be done manually.

claudio525 (commented):
Couple of minor last minute comments. Otherwise looks great to me.

AndrewRidden-Harper (approved):

sungeunbae (dismissed):

claudio525 (approved):

qcore/coordinates.py
```diff
@@ -36,6 +36,8 @@ def wgs_depth_to_nztm(wgs_depth_coordinates: np.ndarray) -> np.ndarray:
     np.ndarray
         An array of shape (N, 3) containing NZTM coordinates (x, y, depth).
     """
+    if wgs_depth_coordinates.shape == (3,):
```
joelridden, line 39: Based on the docstring I would assume this is always the case? Maybe docstring needs to change

qcore/coordinates.py
```diff
@@ -59,8 +61,10 @@ def nztm_to_wgs_depth(nztm_coordinates: np.ndarray) -> np.ndarray:
     np.ndarray
         An array of shape (N, 3) containing WGS84 coordinates (latitude, longitude, depth).
     """
+    if nztm_coordinates.shape == (3,):
```
joelridden, line 64: Same comment as above

qcore/gsf.py
```diff
@@ -109,55 +111,58 @@ def coordinate_meshgrid(
...
-        for i, (length, width, strike, dip, rake, meshgrid) in enumerate(
-            zip(lengths, widths, strikes, dips, rakes, meshgrids)
-        ):
-            strike_step = length / gridpoint_count_in_length(length * 1000, resolution)
-            dip_step = width / gridpoint_count_in_length(width * 1000, resolution)
-            for point in meshgrid:
-                gsf_file_handle.write(
-                    f"{point[1]:11.5f} {point[0]:11.5f} {point[2] / 1000:11.5e} {strike_step:11.5e} {dip_step:11.5e} {strike:6.1f} {dip:6.1f} {rake:6.1f} {-1.0:8.2f} {-1.0:8.2f} {i:3d}\n"
-                )
+        for j in range(n_dip):
```
joelridden, line 152: Not sure how I feel about the variable j. Maybe dip_index or depth_index? As it's the index as you go down the dip / go deeper. Maybe even i could change to plane_index

qcore/gsf.py
```diff
@@ -0,0 +1,216 @@
...
+        - strike
+        - dip
+        - rake
+        - slip (nearly always -1)
+        - init_time (nearly always -1)
+        - seg_no (the fault segment this point belongs to)
+    """
+    with open(gsf_filepath, mode="r", encoding="utf-8") as gsf_file_handle:
+        # we could use pd.read_csv with the skiprows argument, but it's not
+        # as versatile as simply skipping the first n rows with '#'
```
sungeunbae, line 194: Interesting

sungeunbae (dismissed):

AndrewRidden-Harper (approved):

qcore/coordinates.py
lispandfound: No this is to allow code like the following: 

```wgs_depth_to_nztm(np.array([lat, lon, depth]))```

Otherwise you have to do an ugly

```wgs_depth_to_nztm(np.array([lat, lon, depth]).reshape((1, -1)))```

lispandfound: I can update the docstring, however.

joelridden (approved):

claudio525 (approved):
Nice work!

AndrewRidden-Harper (approved):

sungeunbae (approved):

PR #305, opened 2024-05-20
Add oriented bearing function

lispandfound:
Add a function to compute the oriented bearing between two vectors. Not sure how this wasn't already in qcore because it's super useful for finding the strike or dip direction between corners. Here is, for example, how I am using to find the strike direction of a fault given the corners of the fault:

```python
north_direction = np.array([1, 0, 0])
up_direction = np.array([0, 0, 1])
strike_direction = self._corners[1] - self._corners[0]
return qcore.geo.oriented_bearing_wrt_normal(
    north_direction, strike_direction, up_direction
)
```

### DIFF
--- a/qcore/geo.py
+++ b/qcore/geo.py
@@ -155,7 +155,8 @@ def oriented_bearing_wrt_normal(
     """Compute the oriented bearing from u to v with respect to a normal.
 
     This function is useful to calculate, for example, strike and dip
-    directions. The orientation is established via the right hand rule (or refer to the diagram below).
+    directions. The orientation is established via the right hand rule
+    (or refer to the diagram below).
 
            v
            ^
--- a/qcore/geo.py
+++ b/qcore/geo.py
@@ -6,7 +6,7 @@ import functools
 import itertools
 from math import acos, asin, atan, atan2, cos, degrees, pi, radians, sin, sqrt
 from subprocess import PIPE, Popen
-from typing import Tuple, Union, Dict, Any, List
+from typing import Any, Dict, List, Tuple, Union
 from warnings import warn
 
 import numpy as np
@@ -158,7 +158,7 @@ def oriented_bearing_wrt_normal(
     directions. The orientation is established via the right hand rule
     (or refer to the diagram below).
 
-        to_bearing
+        to_direction
            ^
            │
            │
--- a/qcore/geo.py
+++ b/qcore/geo.py
@@ -149,6 +149,56 @@ def ll2gp_multi(
     return xy
 
 
+def oriented_bearing_wrt_normal(
+    u: np.ndarray, v: np.ndarray, normal: np.ndarray
+) -> float:
+    """Compute the oriented bearing from u to v with respect to a normal.
+
+    This function is useful to calculate, for example, strike and dip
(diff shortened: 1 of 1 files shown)


qcore/geo.py
```diff
@@ -149,6 +149,57 @@ def ll2gp_multi(
...
+         ╱
+        ╱
+       ╱
+    normal
+
+    Parameters
+    ----------
+    u : np.ndarray
+    v : np.ndarray
+    normal : np.ndarray
```
joelridden, line 181: A description for these could be great, not sure if there is a better variable name that could be used instead of u and v to better describe what your actually dealing with

lispandfound: Best I can think of is changing u and v to from_direction and to_direction.

joelridden (approved):

sungeunbae (approved):

claudio525 (commented):
Nice addition, only thought i have, does it belong in geo.py? When I think geo.py I think lon/lat

qcore/geo.py
```diff
@@ -149,6 +149,60 @@ def ll2gp_multi(
...
+        The direction to measure the bearing from.
+    to_direction : np.ndarray
+        The direction to measure the bearing to.
+    normal : np.ndarray
+        The normal direction to orient the bearing with via the right hand rule.
+
+    Returns
+    -------
+    float
+        The bearing from u to v with oriented with respect to normal.
```
AndrewRidden-Harper, line 189: I think the first "with" in "The bearing from u to v with oriented with respect to normal." should be deleted

AndrewRidden-Harper (commented):
small typo in docstring

qcore/geo.py
lispandfound: yeah good catch.

lispandfound:
@claudio525 Regarding the relevance of it's inclusion in geo, there is already geometry code in the module that is not related to coordinates. For example, `avg_wbearing`, `angle_diff` (both old functions), and the recently added `closest_points_between_planes`. When I made #303 I put the two coordinate transformations into a separate coordinates module because I didn't think they were "geo" related.

qcore/geo.py
```diff
@@ -149,6 +149,61 @@ def ll2gp_multi(
...
+def oriented_bearing_wrt_normal(
+    from_direction: np.ndarray, to_direction: np.ndarray, normal: np.ndarray
+) -> float:
+    """Compute the oriented bearing between two directions with respect to a normal.
+
+    This function is useful to calculate, for example, strike and dip
+    directions. The orientation is established via the right hand rule
+    (or refer to the diagram below).
+
+        to_bearing
```
sungeunbae, line 161: to_direction?

claudio525 (approved):

joelridden (dismissed):
Sung's comment might be correct just for documentation updates but other than that I'm happy

qcore/geo.py
lispandfound: Ah yes good catch!

AndrewRidden-Harper (approved):

sungeunbae (approved):

sungeunbae:
As @claudio525 and @joelridden previously approved this, I'm merging this

PR #309, opened 2024-06-04
Fix "no module named imp"

### LINKED ISSUE #301: ModuleNotFoundError: No module named 'imp'
- `ModuleNotFoundError: No module named 'imp'`. This module is imported in the `utils` module.  The `imp` module has been removed in Python 3.12, and deprecated since Python 3.4. We should use `importlib` instead.

lispandfound:
Fixes #301 by removing the function `load_py_cfg`. This function seems to be unused according to a [quick search](https://github.com/search?q=org%3Aucgmsim+load_py_cfg&type=code), and it's use of the imp module means you can't use the `utils` module in Python versions >=3.12 (which is required by the gmt code).


lispandfound:
Actually this pull now does a fair bit more. It deletes the following functions:

- `ordered_load`
- `ordered_dump`
- `load_py_cfg`
- `change_file_ext`
- `load_sim_parms`

The first three weren't used outside the utils module, `change_file_ext` is only used in one other place and redefined in gm_hazard. The `load_sim_params` function is only used in slurm_gm_workflow; I will add a PR to move that function into the workflow repo. In addition to the above changes:

- `load_yaml` and `dump_yaml` are now basically aliases to `yaml.safe_load` and `yaml.safe_dump`. I added a note to their docstrings to direct people to just use PyYAML's built-in functionality, but I didn't delete them because it's too much work.
- The `compare_versions` function is now shorter and clearer.
- Every function has numpy-style docstrings and type hints.

sungeunbae (approved):

AndrewRidden-Harper (approved):

claudio525 (approved):

joelridden (approved):

PR #312, opened 2024-06-05
Add grid module

lispandfound:
@joelridden wanted a function that can do what the gsf module does with the meshgrid, but from the usual plane parameters strike, dip direction, dbottom, dtop, etc for some work in nzgmdb. I have broken out the grid functions into a separate module because they aren't quite geometry related, but aren't really related to gsf files any longer.


claudio525 (commented):
Nice!

claudio525 (approved):

joelridden (approved):

PR #314, opened 2024-06-17
Add bounding box + tests + wiki

lispandfound:
This implements the request of @claudio525 in https://github.com/ucgmsim/Pre-processing/pull/230 to make `bounding_box` a qcore module, provide tests and a background wiki explanation. It also adds the github action from https://github.com/ucgmsim/github-wiki-test/pull/1 to qcore in the process of adding the wiki pages.

PR #315, opened 2024-06-17
Cleanup distributions module

lispandfound:
Improves the distributions module so that:

1. Everything is documented and has proper type-signatures
2. Superfluous functions are removed.
3. Everything is mathematically sound.

This pull request is required to move the uncertainty code out of type-5 srf gen in https://github.com/ucgmsim/Pre-processing/pull/230.

The duplication of code in the functions `truncated_weibull` and `truncated_weibull_expected_value` is undesirable, but required at the moment until someone can figure out the source of these magic `c` and `scale` values. I suspect I will need to adjust the shape value `c` for larger `upper_value` thresholds. Will check the textbook and adjust if I spot something helpful.


lispandfound:
the weibull distribution changes will require a couple of tweaks in Pre-processing. I will create a pull-request for these.

claudio525 (approved):
Nice work. In regards to the `c` and `shape` value, you are unsure about the where the specific defaults are coming from?

lispandfound:
Regarding the weibull distribution: the source for the parameter values comes from [here](https://wiki.canterbury.ac.nz/pages/viewpage.action?pageId=119047424). The source data had rescaled the lengths of each fault in the sample to 1, so it's probably not wise to try and change the code to accept an `upper_limit` and remove the duplicated code.

lispandfound:
Also the failing tests in this module are a bug in the tests. Once #314 is merged they'll start working again.

AndrewRidden-Harper (approved):
nice cleanup

joelridden (approved):

PR #316, opened 2024-06-18
Implement source geometry classes

lispandfound:
# Add Source Geometry Definitions

This pull request adds a common source geometry definition for
qcore. Each geometry (Point, Plane, and Fault) minimally contains a
definition of it's boundary and a source-local coordinate system that
can be used to compute all sorts of useful properties.

I intend this to be as general as possible to allow all sorts of
repositories to remove lines of code and make use of the general
source geometry code in this module. **Any suggestions for additions
would be welcome**.


lispandfound:
Closed in favour of the source modelling repo.

PR #320, opened 2024-07-11
Extract data download

joelridden:
Very quick fix to remove downloading data on install and make it optional for people to add if they want to make use of the GMT map data.

Added typer for optional CLI for specifying data location and download from dropbox.


claudio525 (approved):

AndrewRidden-Harper (dismissed):
looks good

download_data.py
```diff
@@ -0,0 +1,50 @@
...
+from urllib.request import urlretrieve
+
+
+PACKAGE_NAME = "qcore"
+PACKAGE_URL = f"https://github.com/ucgmsim/{PACKAGE_NAME}"
+DATA_VERSION = "1.2"
+DATA_NAME = "qcore_resources.tar.xz"
+DATA_URL = f"{PACKAGE_URL}/releases/download/{DATA_VERSION}/{DATA_NAME}"
+
+NO_DATA_ARG = "--no-data"
```
lispandfound, line 13: We can drop the `NO_DATA_ARG` variable now.

download_data.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def get_version(version_path):
+    if os.path.isfile(version_path):
+        return open(version_path).read().strip()
+    else:
+        return None
+
+
+def prepare_data():
```
lispandfound, line 28:
```suggestion
def download_data():
```

download_data.py
```diff
@@ -0,0 +1,50 @@
...
+        print("downloading...")
+        urlretrieve(DATA_URL, DATA_NAME)
+        extract_data(DATA_NAME, PACKAGE_NAME)
+    # final check
+    have_ver = get_version(loc_version)
+    if str(have_ver) != DATA_VERSION:
+        sys.exit("data package issue, please contact repository maintainer")
+
+
+prepare_data()
```
lispandfound, line 50: I would change this to 

```python
if __name__ == '__main__':
    download_data()
```

That way, if I import the module the data would not download until I call `download_data()`

lispandfound (changes requested):
I would like to be able to call the function in this script from another module. I would propose moving this into `qcore/data.py` and then that way I can write code that does 

```python
from qcore import data

data.prepare_data()
```

To download the data in scripts that require it.

lispandfound (approved):

claudio525 (approved):

sungeunbae (approved):

AndrewRidden-Harper (dismissed):

lispandfound (changes requested):
In my own testing of this PR it doesn't quite work as expected. If you call the script with your current working directory anywhere but `qcore/data/` it will fail. To remedy this, I made the following change to the download_data.py script to make things work how it should:

```python
# from pathlib import Path
# import tempfile
# from urllib import request
def download_data():
    download_location = Path(__file__).parent
    loc_version = download_location / "version"

    have_ver = get_version(loc_version)
    if str(have_ver) != DATA_VERSION:
        print("data package missing or incorrect version")
        print("downloading...")
        with tempfile.NamedTemporaryFile() as data_archive_file:
            request.urlretrieve(DATA_URL, data_archive_file.name)
            extract_data(data_archive_file.name, download_location.parent)
    # final check
    have_ver = get_version(loc_version)
    if str(have_ver) != DATA_VERSION:
        sys.exit("data package issue, please contact repository maintainer")
```

Note that the download location is sourced from `__file__`. This approach will not work if you invoke the script from the command line, however. So @joelridden will need to find a different work around to work with both the CLI and the import-and-download methods.

lispandfound:
Also just something to think about, but the GitHub release download host is horrifically slow (like 2Mb/s vs Dropbox at 52Mb/s). Perhaps there is a better place to host nearly 2gb of data than a GitHub release link?

sungeunbae:
> Also just something to think about, but the GitHub release download host is horrifically slow (like 2Mb/s vs Dropbox at 52Mb/s). Perhaps there is a better place to host nearly 2gb of data than a GitHub release link?

Dropbox will do. Just remember this bit!
image

lispandfound:
Link here: https://www.dropbox.com/scl/fi/p5so9irbx43c9g96jr2qr/qcore_resources.tar.xz?rlkey=pgmn69e212c0zx0fauxuikgky&dl=0

claudio525 (approved):

lispandfound (approved):

sungeunbae (approved):

AndrewRidden-Harper (approved):

PR #321, opened 2024-07-18
Improve documentation and raise errors for coords

lispandfound:


qcore/coordinates.py
```diff
@@ -7,7 +7,7 @@
 - wgs_depth_to_nztm(wgs_depth_coordinates: np.ndarray) -> np.ndarray:
     Converts WGS84 coordinates (latitude, longitude, depth) to NZTM coordinates.
 - nztm_to_wgs_depth(nztm_coordinates: np.ndarray) -> np.ndarray:
-    Converts NZTM coordinates (x, y, depth) to WGS84 coordinates.
+    Converts NZTM coordinates (y, x, depth) to WGS84 coordinates.
```
claudio525, line 10: Any particular reason for `y,x` instead of `x,y`?

lispandfound: It's just what pyproj returns. The pyproj library is consistent about this (lat, lon) -> (y, x) and (y, x) -> (lat, lon).

joelridden (approved):

sungeunbae (approved):

qcore/coordinates.py
claudio525: It would be good to keep this consistent with our convention, which I believe is (x,y,depth) and (lon,lat,depth). But we have never defined this properly, so maybe we should briefly chat about this on Wednesday.

qcore/coordinates.py
```diff
@@ -56,7 +62,13 @@ def wgs_depth_to_nztm(wgs_depth_coordinates: np.ndarray) -> np.ndarray:
     array([[5.92021456e+06, 1.75731133e+06, 1.00000000e+02],
            [5.42725716e+06, 1.74893148e+06, 0.00000000e+00]])
     """
-    return np.array(_WGS2NZTM.transform(*wgs_depth_coordinates.T)).T
+    nztm_coordinates = np.array(_WGS2NZTM.transform(*wgs_depth_coordinates.T)).T
+    if not np.all(np.isfinite(nztm_coordinates)):
+        raise ValueError(
+            "Latitude and longitude coordinates given are invalid (did you input lon, lat instead of lat, lon?)"
```
AndrewRidden-Harper, line 68: nice helpful error message

AndrewRidden-Harper (approved):

qcore/coordinates.py
lispandfound: I'll should be in tomorrow (flu permitting) and we can discuss this then. I think latitude, longitude is a better format for a couple of reasons. But the most important is that is recognised by the international standards organisation as the correct order. Quoting [the wikipedia summary of ISO standard 6709](https://en.wikipedia.org/wiki/ISO_6709):

> Order, positive direction, and units of coordinates are supposed to be defined by the CRS. When CRS  identification is missing, the data must be interpreted by the following conventions:
>
> - Latitude comes before longitude
> - North latitude is positive
> - East longitude is positive
> - Fraction of degrees (decimal degrees) is preferred in digital data exchange, while sexagesimal notation is tolerated for compatibility.
```

This standard is why pyproj uses latitude, longitude instead of longitude, latitude. I understand our legacy code (sometimes) uses the longitude, latitude order. But this isn't even consistent in modules like `geo`. My favourite offenders in this category are `ll2gp_multi(coords, mlon, mlat, ...)` vs `gp2ll_multi(coords, mlat, mlon, ...)` (Notice they swap order between two very similar functions).

All my code is (latitude, longitude) in source_modelling, workflow, and nshmdb. I also happen not to use any of the old geometry code. On the other hand, the old code doesn't really use any of my new geometry or coordinate code. Since this coordinate function is heavily used by my code, and not really by old code, I don't think it should conform to the old inconsistent standard.

claudio525: Right wasn't aware that that is the "correct" order, I'm happy to change to that if we all agree on that. Consistency going forward is the key point imo. And yeah legacy code is def not consistent haha.

claudio525 (approved):

PR #322, opened 2024-07-29
Better docstrings

lispandfound:
Continuing my crusade to clean up the old qcore code while I wait for EMOD3D jobs to be taken up on NESI...

Gives docstrings, type annotations and cleans up weird code from the intensity measure and qconfig modules.

### DIFF
--- a/qcore/im.py
+++ b/qcore/im.py
@@ -250,7 +250,7 @@ class IM:
         Parameters
         ----------
         name : str
-            The name of the intensity measure (e.g. "PGV")
+            The name of the intensity measure (e.g. "pSA_1.0")
 
         Returns
         -------
--- a/qcore/im.py
+++ b/qcore/im.py
@@ -1,13 +1,16 @@
 """
-Correct IM column order
-station, component, PGA*, PGV*, CAV*, AI*, Ds*, MMI*, pSA_*, FAS_*, SDI_*
+Module for handling and analyzing intensity measures (IM) in seismic data.
+
+This module provides functionality for working with intensity measures (IMs), which are used to quantify and analyze seismic activity.
 """
 
-from dataclasses import dataclass
 import enum
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Optional, Union
 
-import pandas as pd
 import numpy as np
+import pandas as pd
 
 from qcore import constants
 
@@ -29,6 +32,38 @@ DEFAULT_PATTERN_ORDER = (
 
 
 class IMEnum(constants.ExtendedEnum):
+    """
+    Enum representing different types of intensity measures (IM) used in seismic analysis.
+
+    This enumeration defines a set of standardized intensity measures commonly used to quantify and analyze seismic activity.
+    Each member of this enum corresponds to a specific type of intensity measure.
+
+    Members
+    -------
+    PGA
+        Peak Ground Acceleration, a measure of the maximum acceleration of the ground during an earthquake.
+    PGV
+        Peak Ground Velocity, a measure of the maximum velocity of the ground during an earthquake.
+    CAV
+        Cumulative Absolute Velocity, an intensity measure that accounts for the duration and amplitude of ground shaking.
+    AI
+        Arias Intensity, a measure of the total energy released during an earthquake.
+    Ds575
+        A specific spectral displacement measure with a 0.75-second period.
+    Ds595
+        A specific spectral displacement measure with a 0.95-second period.
+    Ds2080
+        A specific spectral displacement measure with a 2.0-second period.
+    MMI
+        Modified Mercalli Intensity, a qualitative measure of the effects of an earthquake on people and structures.
+    pSA
+        Pseudo Spectral Acceleration, a measure of acceleration response at a specific period.
+    FAS
+        Fourier Amplitude Spectrum, a measure of the amplitude of ground motion as a function of frequency.
+    SDI
+        Spectral Displacement Index, a measure of displacement response at a specific period.
+    """
+
     PGA = enum.auto()
     PGV = enum.auto()
     CAV = enum.auto()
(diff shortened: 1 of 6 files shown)


claudio525 (dismissed):
Great crusade, nice work!

qcore/im.py
```diff
@@ -178,5 +245,17 @@ def get_period_unit(self):
 
     @staticmethod
     def from_im_name(name: str):
+        """Create an IM from a string.
+
+        Parameters
+        ----------
+        name : str
+            The name of the intensity measure (e.g. "PGV")
```
joelridden, line 253: Maybe a better example could be pSA_1.0, just so people get the idea that it also can include the period not just the IM type

joelridden (approved):

joelridden (dismissed):

claudio525 (approved):

lispandfound:
Sorry got extra ambitious and refactored `qcore.shared` as well.

qcore/shared.py
```diff
@@ -1,109 +1,202 @@
...
+    stations : list[str]
+        A list of station names in the file.
+    longitudes : list[float], if locations is True
+        The longitudes of the stations.
+    latitudes : list[float], if locations is True
+        The latitudes of the stations.
+    """
+    with open(station_ffp, "r", encoding="utf-8") as source_file_handle:
+        stations_text = source_file_handle.read()
+        # have to read and replace because pandas only allows one comment
```
claudio525, line 41: Do we actually have commented out lines using `%` in our station files? @sungeunbae @joelridden 
If not, then we should just change this to only support `#`?

claudio525 (commented):
Nice work. Did you test the functions that you changed the code for?

qcore/shared.py
```diff
@@ -1,109 +1,204 @@
...
-        for corner in ["c1=", "c2=", "c3=", "c4="]:
-            for line in lines:
-                if corner in line:
-                    corners.append(list(map(float, line.split()[1:3])))
-                    break
+    with open(model_params_ffp, "r") as model_params_file_handle:
+        corners = {}
+        for line in model_params_file_handle:
+            corner_match = re.match(
+                r"\s+c(\d)=\s+([0-9\.\-\+]+)\s+([0-9\.\-\+]+)", line
```
sungeunbae, line 90: Can we add a comment like 

> matching a string in "c{i}=   {lon}   {lat}" format 
> e.g.)   c1=  172.16403  -41.41359

qcore/shared.py
```diff
@@ -1,109 +1,204 @@
...
+def non_blocking_exe(
+    cmd: Union[str, list[str]],
+    debug: bool = True,
+    stdout: Union[bool, IOBase] = True,
+    stderr: Union[bool, IOBase] = True,
+    **kwargs,
+) -> subprocess.Popen:
+    """Run a command without blocking the calling thread.
+
+    *DO NOT USE THIS FUNCTION* Instead, call subprocess.run or
```
sungeunbae, line 115: I remember this function was developed to help novice python users (in our early days) to easily execute and interact with RG's code to pass arguments. We are probably mature enough to mandate a proper way.

Can you add some risk involved?
Also an example of proper use of subprocess.run will be nice

qcore/shared.py
```diff
@@ -113,32 +208,25 @@ def exe(cmd, debug=True, shell=False, stdout=True, stderr=True, stdin=None, **kw
...
-        return False
-
-    # all tests passed
-    return True
+    Returns
+    -------
+    bool
+        True if the station name is a virtual station identifier.
+    """
+    return bool(re.match(r"[0-9a-f]{7}", station_name))
```
sungeunbae, line 232: love this

qcore/shared.py
joelridden: I don't recall ever seeing a % in these during my time. Probably safe to remove

lispandfound: ok great I'll remove that. It will clean up the code.

qcore/shared.py
lispandfound: yes can do

qcore/shared.py
lispandfound: There isn't a whole lot of risk per-se. The main reason not to use it in my opinion is that it provides no real advantage `subprocess.check_call`. I mean the difference between

```python
>>> exe('ls')
# Some popen object
``` 
and 
```python
>>> subprocess.check_output(['ls'])
"file1.py\nfile2.py\n..."
```
is pretty neglible. Arguably the second is more noob friendly because it returns the stdout immediately and this is usually what they want.

I suppose there would be a risk that you might leak the popen object if you forget to wrap it in a context manager, which you do not need to care about for `run`, `check_call`, and friends.

qcore/im.py
```diff
@@ -111,22 +155,42 @@ def order_ims(unsorted_ims, pattern_order=DEFAULT_PATTERN_ORDER):
                 sorted_indices = np.argsort([len(im) for im in cur_ims])
 
             # Sort the columns names
-            adj_ims = adj_ims + list(np.asarray(cur_ims)[sorted_indices])
+            adj_ims.extend(np.asarray(cur_ims)[sorted_indices])
     # Deal with columns that aren't handled by the pattern.
     # These are just added to the end, in the original order
-    if len(adj_ims) != len(unsorted_ims):
-        [adj_ims.append(im) for im in unsorted_ims if im not in adj_ims]
+    adj_ims.extend(im for im in df.columns if im not in adj_ims)
```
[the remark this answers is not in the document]
lispandfound: Yup! `extend` will accept any iterator as argument, you do not need to make the garbage collector work harder destroying the array you just made.

joelridden (dismissed):
Not sure why the tests are complaining about pandas, bit weird but should be fixable

lispandfound:
> Not sure why the tests are complaining about pandas, bit weird but should be fixable

Because, for some reason, pandas is not a dependency in qcore at this stage. I just added it.

AndrewRidden-Harper (approved):
looks good

joelridden (approved):

claudio525 (approved):

sungeunbae (approved):

PR #324, opened 2024-08-07
Updating rrup test

sungeunbae:
Moved  this test 
https://github.com/ucgmsim/IM_calculation/blob/0dac6f172ddbb11e9b4faadf2e3ab27746e8f0d3/IM_calculation/test/test_source_site_dist/test_src_site_dist_calc.py

out of IM_calculation repo.

IM_calculation test has a separate test_common_set_up.py that provides some utility functions including data download and directory setting up. 

I felt copying this to enable this test in qcore repo is more than necessary. IM calculation tests have something in common. When they share the same data set, so one-time downloading the data zip file from external host and setting things up for the entire test set is justified. However, qcore tests are more individualized and each existing test owns its own test data folder. Making a single data zip file that contains all these tests of different nature add unnecessary complexity to the test design and maintenance.

Hence I placed the sample input and output data under test_source_site_dist folder, like other qcore tests. The sample data is tiny (1.3M), and won't add too much overhead for most users cloning qcore repo.


qcore/test/test_source_site_dist/test_src_site_dist_calc.py
```diff
@@ -0,0 +1,124 @@
...
+    srf_header = pickle.load(
+        open(os.path.join(root_path, INPUT, function + "_srf_header.P"), "rb")
+    )
+    locations = np.load(os.path.join(root_path, INPUT, function + "_locations.npy"))
+
+    out_rx = np.load(os.path.join(root_path, OUTPUT, function + "_rx.npy"))
+    out_ry = np.asarray([-711.41646299])
+    # below is the original line.
+    # replaced by https://github.com/ucgmsim/IM_calculation/commit/c5ff32c10a85c23d6503bdaebdf4dfaaf2ee6bf8
+    # out_ry = np.load(os.path.join(root_path, OUTPUT, function + "_ry.npy"))
```
joelridden, line 119: I wonder if we should check this dodgy test patch, or at least change the value stored in the .npy to reflect this value and load it instead

sungeunbae: Good advice. Updated the dodgy .npy file :)

joelridden (approved):

lispandfound (approved):

claudio525 (approved):

PR #328, opened 2024-09-04
Optional[A | B] is not supported in old Python

sungeunbae:
We have Python 3.9 on Maui and 3.7 on Nurion.

Optional[A | B]  should be Optional[Union[A,B]] while we need to support these machines..


lispandfound (approved):
At some point we should add a qcore typing module that adds common-sense type variables, like Union[Path, str] that are used all over the place. To avoid issues like this.

joelridden (approved):

claudio525 (approved):

PR #330, opened 2024-09-11
Add Bearing Correction Functions for NZTM and Great-circle Geometry

lispandfound:
I have added two functions to the coordinates module, which translates bearings between great-circle geometry bearings and nztm bearing (and vice-versa).

# Why do we need this?

We have two ways of computing bearings between points:

1. Using `geo.ll_bearing`, which utilises great-circle geometry (assuming the earth is a sphere) and,
2. Using the `coordinates` module to transform lat-lon into NZTM coordinates, and computing the bearing using standard euclidean geometry.

The issue is that these two methods do not agree entirely because NZTM gives a different (more accurate, for New Zealand) model for the surface of the Earth than great-circle geometry because the earth is ellipsoidal. This means starting at (-43.538, 172.6474) and travelling at a 45 degree bearing for 100km using `geo.ll_shift` is different to converting that same coordinate to NZTM and then adding $1e5 * (1/\sqrt{2}, 1/\sqrt{2})$ to travel at a 45 degree bearing for 100 km. The difference in final destination for that example will be about a kilometre.

What we need here is a correction, a function that can compute the angle that gets the NZTM bearing as close to the great-circle bearing as possible. This correction computation is the point of the PR.

The correction required depends on the distance travelled, the initial bearing, and most importantly on the origin point. The difference in bearing is largest in my testing when the origin is in Fiordland and near Auckland. Here the bearing correction can be up to 3 or 4 degrees. **See the comments for an example of how big the effect can be**.

# What Have We Added?

We have added two functions that translate bearings between great-circle geometry:

1. `nztm_bearing_to_great_circle_bearing` which computes the great-circle equivalent bearing for an NZTM bearing,
2. `great_circle_bearing_to_nztm_bearing` which computes the NZTM equivalent bearing for a great-circle bearing.

### DIFF
--- a/qcore/coordinates.py
+++ b/qcore/coordinates.py
@@ -22,6 +22,8 @@ from typing import Union
 import numpy as np
 import pyproj
 
+from qcore import geo
+
 # Module level conversion constants for internal use only.
 _WGS_CODE = 4326
 _NZTM_CODE = 2193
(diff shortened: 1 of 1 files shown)


lispandfound:
The need for this is clearest with the velocity models. Consider the Swedge1 simulation from the `v24p6` Cybershake run. Reading the great-circle bearing from the `vm_params` yaml file and passing that to the `bounding_box` module (which assumes input bearing is an NZTM bearing) yields an incorrect bounding box. This is easily seen if we plot the top-level velocity model density over the bounding box definition because the land doesn't line up with the underlying plot.

output

The yaml file has an uncorrected bearing of ~42 degrees. The nztm bearing from `ll_bearing_correction` is `38.5` degrees. Using the corrected bearing we get a picture like the following.

corrected-output

Notice that the corrected bearing results in the bounding box corners in NZTM making sense (the land in the plot lines up with the land in the velocity model).

joelridden (approved):
Great PR description, really clear

sungeunbae (approved):
Excellent

AndrewRidden-Harper (approved):
very nice

PR #331, opened 2024-09-11
Support for Mahuika Milan nodes 

sungeunbae:
Mahuika Milan nodes' hostname is something like "wmc005".
When an environment is activated, it will check the hostname via platform.node() call to determine if this is Maui or Mahuika (or Nurion) then loads the correct config file (eg. machine_mahuika.json vs machine_maui.json ) 

This fix ensures the correct config file is loaded for a milan node.

### DIFF
--- a/qcore/config.py
+++ b/qcore/config.py
@@ -27,7 +27,7 @@ from enum import Enum, auto
 from json import load
 from pathlib import Path
 from platform import node
-from typing import Optional, Tuple, TypedDict
+from typing import Optional, Tuple, TypedDict, Union
 
 
 class ConfigDict(TypedDict):
@@ -43,7 +43,7 @@ class ConfigDict(TypedDict):
 
 MACHINE_MAPPINGS = {
     r"ni\d{4}|maui.*": "maui",
-    r"wb\d{4}|mahuika.*": "mahuika",
+    r"wb\d{4}|wmc\d{3}|mahuika.*": "mahuika",
     r".*stampede.*": "stampede2",
     r"(login|node|nurion).*": "nurion",
 }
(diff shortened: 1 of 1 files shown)


sungeunbae:
Possibly other Mahuika partitions (eg. hugemem) may need a similar treatment.

sungeunbae:
I guess other Mahuika partitions (eg. hugemem) may need a similar treatment.

claudio525 (approved):

joelridden (dismissed):

lispandfound (approved):

lispandfound (changes requested):
Could you add the hugemem node support now so we don't have to add it later, if it is just a simple regex change?

sungeunbae:
> Could you add the hugemem node support now so we don't have to add it later, if it is just a simple regex change?

There are many such partitions: As the node name patterns are not published, I just collected from squeue output, which may not be complete.

image

```
(python3_mahuika) sjn87@mahuika01: /nesi/project/nesi00213/Environments/sarah2024/qcore/qcore$ squeue |grep milan
50100412      joinsu   landcare0 ppa00        128    500G milan   Sep 11 21:16     1:32:58 RUNNING  wml067              
50100591      xwan754  uoa03263  t1209_H2CC    64     32G milan   Sep 11 21:12    22:27:56 RUNNING  wmc027              
50100618      xwan754  uoa03263  t1209_H2CC    64     32G milan   Sep 11 21:15    22:30:19 RUNNING  wmc032              

```

joelridden (approved):

AndrewRidden-Harper (approved):

lispandfound (approved):
Great!

PR #333, opened 2024-10-01
Validate vm bounds

sungeunbae:
`validate_vm_bounds()` is called during the vm_params step, specifically by `rel2vm_params.py`. 

This function was originally using longitude and latitude to check inside/outside check, but it is safer to convert to the NZTM coordinates first. 

https://github.com/ucgmsim/Pre-processing/pull/238

I also found the code was not particularly efficient, as it suffices to test the srf corner points. Hence this PR.

### DIFF
--- a/qcore/validate_vm.py
+++ b/qcore/validate_vm.py
@@ -107,7 +107,7 @@ def validate_vm_params(vm_params: str, srf: str = None):
     if srf is not None:
         srf_bounds = get_bounds(srf)
     else:
-        srf_bounds = None
+        srf_bounds = []
     errors.extend(validate_vm_bounds(polygon, srf_bounds))
 
     if errors:
@@ -177,7 +177,7 @@ def validate_vm_files(vm_dir: str, srf: str = None):
         if srf is not None:
             srf_bounds = get_bounds(srf)
         else:
-            srf_bounds = None
+            srf_bounds = []
         errors.extend(validate_vm_bounds(polygon, srf_bounds))
 
     if errors:
--- a/qcore/validate_vm.py
+++ b/qcore/validate_vm.py
@@ -225,23 +225,22 @@ def validate_vm_bounds(polygon: npt.ArrayLike, srf_bounds: npt.ArrayLike):
     """
 
     errors = []
+    assert len(srf_bounds) > 0, "No srf bounds given"
 
-    if len(srf_bounds) > 0:
-        vm_polygon = Polygon(
-            coordinates.wgs_depth_to_nztm(
-                np.array(polygon)[:, [1, 0]]
-            )  # [lon, lat] to [lat, lon]
-        )
+    # Check if the SRF domain is within the VM domain
+    vm_polygon = Polygon(
+        coordinates.wgs_depth_to_nztm(
+            np.array(polygon)[:, [1, 0]]
+        )  # [lon, lat] to [lat, lon]
+    )
 
-        all_inside = all(
-            vm_polygon.contains(
-                Point(coordinates.wgs_depth_to_nztm(np.array([lat, lon])))
-            )
-            for plane in srf_bounds
-            for lon, lat in plane
-        )
-        if not all_inside:
-            errors.append("Srf extents not contained within velocity model corners")
+    all_inside = all(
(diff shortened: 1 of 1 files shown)


qcore/validate_vm.py
```diff
@@ -181,37 +185,47 @@ def validate_vm_files(vm_dir: str, srf: str = None):
     return True, ""
 
 
-def validate_vm_bounds(polygon, srf_bounds=None):
+def validate_vm_bounds(
+    polygon: npt.ArrayLike, srf_bounds: Optional[npt.ArrayLike] = []
```
joelridden, line 189: Is this function ever actually used in the case that srf_bounds is not needed? And only checks against the DEM / do we want to make sure that we always do check with the srf_bounds by making it not optional

joelridden (commented):
Looks good, just had 1 thought that could maybe remove some logic

qcore/validate_vm.py
sungeunbae: Sometimes validate_vm_bounds() is called with no srf.  This function is called from `validate_vm_params()` and `validate_vm_files()`  which are invokved if validate_vm.py is used as a script (at the end of VM generation step in the workflow. We don't provide SRF to run validate_vm.py. See [vm_gen.sl](https://github.com/ucgmsim/slurm_gm_workflow/blob/ba7aaccd74c5633e28c5d7865978bc0df7e2899b/workflow/automation/org/nesi/vm_gen.sl#L51)

joelridden: From what I can see an SRF is a required parameter of vm_gen.sl but not used? Why would we not test the srf bounds, would this reduce the potential issues that were there for the subduction runs? I could be wrong but seems like it'd be good to actually test with this (I know it's a little out of scope for this PR)

sungeunbae: Good point, I knew there were a few cases where it has no srf, without looking at the bigger picture. Indeed, as you pointed out, vm_gen.sl can use SRF it already has to call validate_vm.py.  Good idea to fix it there (it is a different repo)

This is a better excuse: https://github.com/ucgmsim/Pre-processing/blob/a512a1866f3edd6425a2b68d9cf96ce5a39f351c/VM/vm_params2vm.py#L106
it calls validate_vm_files() with no SRF.

As we have just seen this, this bound is checked at multiple places, maybe repeatedly. Let's review when.

1. rel2vm_params.py: with SRF (It also calls plot_vm() with SRF, but plot_vm() doesn't do bound check)
2. vm_params2vm.py:  without SRF. 
3. vm_gen.sl : calling validate_vm.py as a script without SRF (but we will fix it to use SRF)
4. plot_vm.py:  Only when called as a script.  This [PR](https://github.com/ucgmsim/Pre-processing/pull/238 ) will enforce SRF and it needs this PR merged too.

Note that 2 and 3 are essentially the same step, but doing the same check twice (if done though the workflow). 
I had a domain issue with the subduction faults because I was re-using the vm_params produced elsewhere. If vm_gen.sl utlized SRF, we should have detected an issue during 3. 

vm_params step (1 above) and vm_gen step (2/3 above) are usually done sequentially, so this bound check may seem repetitive and redundant.  As things are not always done in nice order (eg. using vm_params made somewhere else), let's keep the check all places.

Now, your suggestion @joelridden . I think it is a good idea to make SRF mandatory here, but have a separate function that checks aganst DEM to accommodate 2 above.

joelridden: Thank you for the detailed information, very much appreciated! Totally agree

lispandfound (approved):

joelridden (approved):
Great! Does there need to be any updates to the code elsewhere for just validate_region()? since there was a rename / split of the validate_vm_bounds functions

sungeunbae:
> Great! Does there need to be any updates to the code elsewhere for just validate_region()? since there was a rename / split of the validate_vm_bounds functions

`validate_region()` is used in a couple of places in this Pre-processing PR: https://github.com/ucgmsim/Pre-processing/pull/238

AndrewRidden-Harper (approved):

sungeunbae:
@claudio525  Any comment, please?

claudio525 (approved):

PR #302, opened 2024-05-09
Custom filename

joelridden:
Allows a custom filename for writing a text file
Utilized by the NZGMDB for ensuring the outputs are not overwritten since Station.Comp can be common for multiple records

PR #334, opened 2024-10-14
seis2txt() was renamed to timeseries_to_text() but still referenced

sungeunbae:
seis2txt() function was renamed to timeseries_to_text() with minor alterations (https://github.com/ucgmsim/qcore/pull/302/files), but we missed 3 places that still reference the old function name.

It was also noticed that the "prefix" parameter of `all2txt()` has a room for improvement. `prefix` can be a directory path plus a file name prefix. If the prefix ends with /, "file name prefix" part can be empty, and the entire prefix becomes the directory path.
If the prefix doesn't end with /, the last bit becomes the file name prefix, and the remaining part becomes the directory path. We don't handle these cases separately, or don't make sure the directory is available, so potentially the function call fails or writes files in an unexpected location with unintended file names.

I believe this PR resolves the issue.

### DIFF
--- a/qcore/timeseries.py
+++ b/qcore/timeseries.py
@@ -515,7 +515,7 @@ class LFSeis:
                 title=title,
             )
 
-    def all2txt(self, prefix="./", dt=None, f="vel"):
+    def all2txt(self, prefix: str = "./", dt: float = None, f: str = "vel"):
         """
         Creates waveforms in text files for all stations.
         Note: This function is not designed to be used other than for single/debug use.
@@ -715,7 +715,7 @@ class HFSeis:
                 title=title,
             )
 
-    def all2txt(self, prefix="./", dt=None):
+    def all2txt(self, prefix: str = "./", dt: float = None):
         """
         Creates waveforms in text files for all stations.
 
@@ -881,7 +881,7 @@ class BBSeis:
         if prefix is None:
             return xyz
 
-    def all2txt(self, prefix="./", f="acc"):
+    def all2txt(self, prefix: str = "./", f: str = "acc"):
         """
         Extracts waveform data from the binary file and produces output in text format.
         Note: For compatibility. Should run slices in parallel for performance.
--- a/qcore/timeseries.py
+++ b/qcore/timeseries.py
@@ -528,13 +528,12 @@ class LFSeis:
         ----------
         prefix : str, optional, default="./"
             The prefix is an output path combined with an optional filename prefix.
-            If prefix ends with /, its prefix_filename is empty, otherwise, prefix_filename is the last part of the prefix
             eg. prefix = "dir1/dir2/XXX", prefix_filename = "XXX" and prefix_dirname = "dir1/dir2"
             eg. prefix = "dir1/dir2/", prefix_filename = "" and prefix_dirname = "dir1/dir2"
-        dt : float, optional
-            The time step of the data, by default None
-        f : str, optional
-            The type of data to save, by default "acc". Options are "acc" and "vel"
+        dt : float, optional, default=None
+            The time step of the data
+        f : str, optional, default="vel"
+            The type of data to save. Options are "acc" and "vel"
         """
 
         if dt is None:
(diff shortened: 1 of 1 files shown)


qcore/timeseries.py
```diff
@@ -514,17 +517,35 @@ def vel2txt(self, station, prefix="./", title="", dt=None, acc=False):
 
     def all2txt(self, prefix="./", dt=None, f="vel"):
```
joelridden, line 518: Might as well typehint the function too

qcore/timeseries.py
```diff
@@ -696,14 +717,30 @@ def acc2txt(self, station, prefix="./", title="", dt=None):
 
     def all2txt(self, prefix="./", dt=None):
```
joelridden, line 718: Same with this one

qcore/timeseries.py
```diff
@@ -846,12 +883,25 @@ def save_txt(self, station, prefix="./", title="", f="acc"):
```
joelridden, line 883: And here

joelridden (approved):

qcore/timeseries.py
```diff
@@ -502,29 +505,47 @@ def vel2txt(self, station, prefix="./", title="", dt=None, acc=False):
...
+
         if dt is None:
             dt = self.dt
         acc = f == "acc"
+
+        prefix_chunks = prefix.split("/")
+        prefix_filename = prefix_chunks[-1]
+        prefix_dirname = Path("/".join(prefix_chunks[:-1])).resolve()
+        prefix_dirname.mkdir(parents=True, exist_ok=True)
         for s in self.stations.name:
```
lispandfound, line 547: I'd simplify this to something like:

```suggestion
        prefix_path = Path(prefix)
        prefix_path.parent.mkdir(parents=True, exist_ok=True)
        for s in self.stations.name:
            self.vel2txt(s, prefix=prefix, title=prefix_path.stem, dt=dt, acc=acc)
```

lispandfound: You may have to `resolve` there again idk.

qcore/timeseries.py
```diff
@@ -683,27 +704,43 @@ def acc2txt(self, station, prefix="./", title="", dt=None):
...
         if dt is None:
             dt = self.dt
+
+        prefix_chunks = prefix.split("/")
+        prefix_filename = prefix_chunks[-1]
+        prefix_dirname = Path("/".join(prefix_chunks[:-1])).resolve()
+        prefix_dirname.mkdir(parents=True, exist_ok=True)
         for s in self.stations.name:
-            self.acc2txt(s, prefix=prefix, title=prefix, dt=dt)
+            self.acc2txt(s, prefix=prefix, title=prefix_filename, dt=dt)
```
lispandfound, line 743: similarly here.

qcore/timeseries.py
```diff
@@ -758,7 +795,7 @@ def __init__(self, bb_path):
         # read header - strings
         self.lf_dir, self.lf_vm, self.hf_file = np.fromfile(
             bbf, count=3, dtype="|S256"
-        ).astype(np.str_)
+        ).astype(np.unicode_)
```
lispandfound, line 799: `np.unicode_` is an alias of `np.str_` [according to the docs](https://numpy.org/doc/1.25/reference/arrays.scalars.html#numpy.unicode_). Is there any reason to use that?

qcore/timeseries.py
```diff
@@ -844,14 +881,27 @@ def save_txt(self, station, prefix="./", title="", f="acc"):
...
+            The prefix is an output path combined with an optional filename prefix.
+            eg. prefix = "dir1/dir2/XXX", prefix_filename = "XXX" and prefix_dirname = "dir1/dir2"
+            eg. prefix = "dir1/dir2/", prefix_filename = "" and prefix_dirname = "dir1/dir2"
+        f : str, optional
+            The type of data to save, by default "acc". Options are "acc" and "vel"
+        """
+        prefix_chunks = prefix.split("/")
+        prefix_filename = prefix_chunks[-1]
+        prefix_dirname = Path("/".join(prefix_chunks[:-1])).resolve()
+        prefix_dirname.mkdir(parents=True, exist_ok=True)
```
lispandfound, line 902: Similarly here.

lispandfound (changes requested):

AndrewRidden-Harper (approved):

qcore/timeseries.py
sungeunbae: It is slightly trickier.
Consider prefix "AA/BB/CC" vs "AA/BB/CC/".
The former means in AA/BB directory, files with CC prefix eg. CC0000, CC0001...
The latter means in AA/BB/CC directory, files with empty prefix, so 0000, 0001 ...
Your suggested change will treat both cases as the former case.

qcore/timeseries.py
sungeunbae: This change somehow slipped in: https://github.com/ucgmsim/qcore/commit/20de500cc3e06758e6919056b182cfa8c4b21909

claudio525 (approved):

lispandfound (approved):

PR #335, opened 2024-10-16
New Workflow Changes

lispandfound:
Small changes for the new workflow:

1. Add a coordinate patchgrid function to `qcore.grid`. This function is used to generate geometry for SRFs. The difference between this and  a coordinate grid is easily summarised in the following image:
image
2. Add the mag scaling packages to `setup.py`. This made them easier to use in workflow and source modelling

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ from setuptools import setup
 setup(
     name="qcore",
     version="1.2",
-    packages=["qcore", "qcore.uncertainties", "qcore.uncertainties.magnitude_scaling"],
+    packages=["qcore"],
     url="https://github.com/ucgmsim/qcore",
     description="QuakeCoRE Library",
     package_data={
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ from setuptools import setup
 setup(
     name="qcore",
     version="1.2",
-    packages=["qcore"],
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -5,7 +5,7 @@
 setup(
     name="qcore",
     version="1.2",
-    packages=["qcore"],
+    packages=["qcore", "qcore.uncertainties", "qcore.uncertainties.magnitude_scaling"],
```
joelridden, line 8: make a __init__.py instead for these

joelridden (commented):
We should at some point write tests for grid.py?

sungeunbae:
> Small changes for the new workflow:
> 
> 1. Add a coordinate patchgrid function to `qcore.grid`. This function is used to generate geometry for SRFs. The difference between this and  a coordinate grid is easily summarised in the following image:
>    image
 
 Meaning a grid made of centroids?

sungeunbae (dismissed):

joelridden (dismissed):

lispandfound:
> We should at some point write tests for grid.py?

This is a good idea. I'll include some now.

joelridden (approved):

AndrewRidden-Harper (approved):

PR #339, opened 2024-11-20
Add the ability to compute point -> line segment distances

lispandfound:
Adds a new function `geo.point_to_segment_distance` which computes the closest distance between a point and a line segment. Required by https://github.com/ucgmsim/source_modelling/pull/13 for the rrup distance refactor.

### DIFF
--- a/qcore/geo.py
+++ b/qcore/geo.py
@@ -1258,11 +1258,11 @@ def point_to_segment_distance(
     Parameters
     ----------
     p : npt.ArrayLike
-        The first point of the line segment.
+        A point to measure distance to.
     q : npt.ArrayLike
-        The second point of the line segment.
+        The first point of the line segment.
     r : npt.ArrayLike
-        A third point to measure distance to.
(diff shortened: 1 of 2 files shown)


qcore/geo.py
```diff
@@ -1247,3 +1248,40 @@ def rotation_matrix(angle: float) -> np.ndarray:
         The 2x2 rotation matrix.
     """
     return np.array([[np.cos(angle), -np.sin(angle)], [np.sin(angle), np.cos(angle)]])
+
+
+def point_to_segment_distance(
+    p: npt.ArrayLike, q: npt.ArrayLike, r: npt.ArrayLike
+) -> float:
+    """Compute the shortest distance between a point and a line segment.
```
claudio525, line 1256: Nice simplification! I think it would be good to add a link to the math exchange question that this is based on. It provides lots of details for the non-fluent math users.

joelridden (approved):

claudio525 (approved):

PR #341, opened 2024-12-20
Scientific Data Registry

lispandfound:
# The Problem
We have a large number of big files we use in simulations and processing. A small sample includes:

1. The NSHM rupture database, 
2. VS30 and lat/lon for stations,
3. The (C|S)PT database,
4. 1D velocity profiles,
5. 2D velocity model basins.

All these files have multiple versions. We also pick certain versions (not necessarily the latest version) for Cybershake and validation runs.  We need to download these files on-site at UC, in NeSI HPC clusters, and on work laptops from home. We want a centralised registry of these scientific files that has:

1. Multiple versions for the same file. E.g. we have station lat, lon and vs30 data but many versions `gns_2023`, `non_uniform_fd_400`, etc.
2. A way to bless certain versions as fit for validation or Cybershake. PhD researchers should be able to point at these versions and be sure they are using the best practice files for their workflows.
3. An easy-to-use interface that makes switching versions and validating downloads easy.

This PR is an attempt to solve this problem with the development of a file registry for scientific data.

# The Architecture

image

The registry consists of three components:

1. An HTTP Server hosting the raw files (more on this in a second),
2. A registry containing versioned downloadable file names and hashes for those files,
3. A local cache for the user to store downloaded files.

The purpose of the registry is to validate downloaded files match what the user expects. It also ensures that if the user writes code against the latest Cybershake data files with checksum `a`, and then the data updates a year later to have checksum `b`, then the user is warned that the data no longer matches their expectations and they can review the changes (or ask someone who knows).

## Wait. Why not use Git?

Why not indeed. It turns out git is not suitable for our purposes because the files we store are too large. The latest (S)CPT database is 6GB, the NSHM db is 600Mb, the velocity is 6 or 7 gbs total, and Git LFS is fundamentally too slow (and usage limited on GitHub) for our purposes. There really aren't any off the shelf solutions for this problem that make sense for our workflow (believe me, I really tried looking for them!). So instead, we have this somewhat bespoke solution that still maintains the key advantage of a git repo: **all updates to the data registry must go through a pull request before they can be used in any code**.  

## An Example of the Workflow

Let's say we are only trying to keep track of a few versions of a file `science_data.txt`. This file, appropriately, contains mission critical scientific data. We created this file in 2023, and updated it in 2024. The structure on the HTTP server looks like

```
.
├── 2023
│   └── science_data.txt
├── 2024
│   └── science_data.txt
└── latest
    └── science_data.txt -> 2024/science_data.txt
```

The `latest` version just points to the latest good version of the file (we would probably have two such tags, `cybershake` and `validation`). Additionally, we have checksums for these files

```
$ sha256sum **/science_data.txt 
55415e3d094b6dde916f6bcf8b6f4bb3fdbdeed92c091adc4226ae5040f4e599  2023/science_data.txt
b5ac62f8f1958ec5055b0e8004d16e8df25e70ca776e668c8ca66b2e019439f7  2024/science_data.txt
``` 

In `qcore/registry/registry.txt` we have a space-separated list of these files and their checksums. These define the hashes we expect these files to have.

```
latest/science_data.txt sha256:b5ac62f8f1958ec5055b0e8004d16e8df25e70ca776e668c8ca66b2e019439f7
2024/science_data.txt sha256:b5ac62f8f1958ec5055b0e8004d16e8df25e70ca776e668c8ca66b2e019439f7
2023/science_data.txt sha256:55415e3d094b6dde916f6bcf8b6f4bb3fdbdeed92c091adc4226ae5040f4e599
```

Then researcher Alice can download this all-important data and read it using the `qcore.registry.files` module.

```python
>>> from qcore.registry import files
>>> files.get_file('science_data.txt', version='latest').read_text() # get_file returns a path to the downloaded file on disk. 
'hello this is some scientific data\n'
>>> files.get_file('science_data.txt', '2023').read_text()
'older version of the file!\n'
```

In the background this `get_file` function uses the [pooch](https://www.fatiando.org/pooch/latest/) library to download, validate, and cache files from the server. 

> Pooch is a Python library that can manage data by downloading files from a server (only when needed) and storing them locally in a data cache (a folder on your computer).
> - Pure Python and minimal dependencies.
> - Download files over HTTP, FTP, and from data repositories like Zenodo and figshare.
> - Built-in post-processors to unzip/decompress the data after download.
> - Designed to be extended: create custom downloaders and post-processors.
> Are you a scientist or researcher? Pooch can help you too!
> - Host your data on a repository and download using the DOI.
> - Automatically download data using code instead of telling colleagues to do it themselves.
> - Make sure everyone running the code has the same version of the data files.

## Updating the Registry

Now let's say researcher Alice wants to update the registry with a shiny new version of the science data containing an important update. The first thing she does is upload her new science data to the server.

```
$ cat 2025/science_data.txt
Shiny new data oh my!
```

She also updates the symlink `latest/science_data.txt` to point at `2025/science_data.txt`. 

```
.
├── 2023
│   └── science_data.txt
├── 2024
│   └── science_data.txt
├── 2025
│   └── science_data.txt
└── latest
    └── science_data.txt -> 2025/science_data.txt
```

But this isn't enough. In fact if Alice tries to download this data on her computer she gets a nasty error!

```
>>> files.get_file('science_data.txt', 'latest').read_text()
ValueError: SHA256 hash of downloaded file (science_data.txt) does not match the known hash: expected sha256:b5ac62f8f1958ec5055b0e8004d16e8df25e70ca776e668c8ca66b2e019439f6 but got 3797da223e1f783cbac35baa9dba738a8e28f1ce2a3fd68b6224dee37fbac18c. Deleted download for safety. The downloaded file may have been corrupted or the known hash may be outdated.
```

This informs her that the file she tried to download doesn't match the registry in QCore, and that the downloaded file was deleted in case it is corrupted or malicious. At this point Alice talks to developer Bob who informs her she must validate the files by **creating a PR on the QCore repository**. This is how we can achieve a review workflow for scientific data without tracking individual files in Git. Alice creates a PR to QCore that simply changes the `registry.txt` file

```
diff on qcore/registry/registry.txt
~ latest/science_data.txt sha256:3797da223e1f783cbac35baa9dba738a8e28f1ce2a3fd68b6224dee37fbac18c
+ 2025/science_data.txt sha256:3797da223e1f783cbac35baa9dba738a8e28f1ce2a3fd68b6224dee37fbac18c
```

and then through the GitHub PR system we can review the changes. In reality, we'd do this the other way around: updating the registry in QCore first, and then developer Bob with write permissions on the HTTP server would update the scientific data himself after the PR merges. I just wanted to illustrate how the validation works.


lispandfound:
I have changed the registry and moved it to a git-based solution. My commentary on the git-based solution I think is still valid, and we'll wait to see if it becomes slow. However, I have done some more review and I think that it might be ok for our purposes. The problems really arise when you have large files that are edited very often (due to the cost of duplicating these files in the repo). I don't think this is very likely for files in this repository. Moreover, I think it's worth trying git large file system (LFS) first because besides concerns of file sizes it does really neatly solve the problem of reviewing and versioning files.

# How the Git-based Solution Works
The Git-based solution versions files using commits, tags, or branches. The [registry repo](https://github.com/ucgmsim/registry/tree/dev/) contains a  `registry.txt` file with paths and hashes of those files (used to check downloads completed successfully), and paths to files. To get the scientific defaults for the commit `e0d4a7554c35b266340cd5b81fdac8c20d33798d` (which is where the registry repo link points to), you simply build the registry using that commit as the version.

```python
>>> from pathlib import Path
>>> from qcore import registry
>>> pooch = registry.qcore_registry(version = 'e0d4a7554c35b266340cd5b81fdac8c20d33798d')
>>> print(Path(pooch.get_file('v24_2_2_4/defaults.yaml')).read_text())
emod3d:
  all_in_one: 1
  bfilt: 4
  bforce: 0
  dampwidth: 0
  dblcpl: 0
  dmodfile: rho3dfile.d
  dtts: 20
  dump_itinc: 4000
  dxout: 1
# and so on.. 
``` 
If you do not supply a version it will get the latest commit on the `main` branch and use that as the version. Note that because the repo contains the `registry.txt` file the pooch registry will update files that change between usages.

# Updates and Approvals Process
Now the approvals process is pretty familiar. The `main` branch contains the good copy of these files, and is what researchers can point to without thinking about whether the files are good. If a researcher wants to make a change to the `main` branch, they create a new branch make an edit and then PR their changes when they are done. They can use their changes in the mean time for their own work, they can select the branch they work on as the `version` parameter to `qcore_registry`. There will be a GitHub workflow that verifies that the files have hashes that matches what is in the `registry.txt` file and suggests changes so researchers don't have to learn how to generate these themselves.

lispandfound:
You can see an example adding the two VS30 files talked about to the registry: https://github.com/ucgmsim/registry/pull/1.

qcore/registry.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Creates and returns a Pooch registry object for managing file downloads.
+
+    Parameters
+    ----------
+    cache_path : Path, optional
+        The directory where files will be cached. Defaults to the OS cache
+        directory for "quakecore".
+    registry : dict[str, str], optional
+        Registry database to use. Consists of
```
joelridden, line 39: Just missing version in the docstring

joelridden (commented):
Love the solution. Hopefully we don't run into too many issues with LFS

joelridden (approved):

lispandfound:
I have added a feature to request a file using `fetch_file`. This function is intended to safely access files in parallel by acquiring a lock on them if pooch would download or update their contents.  The reason for this function is that, in the workflow, we spawn a lot of processes like `nshm-to-realisation` that need to access the NSHM database to source geometry. If each of these called `pooch.fetch` in separate processes then they would each separately initiate a download of the NSHM database. This is best illustrated in the following diagram

```mermaid
sequenceDiagram
    participant Process1 as Process 1
    participant Process2 as Process 2
    participant Server as File Server
    participant File as Local File

    Process1->>File: Check if file exists
    File-->>Process1: File not found
    Process1->>Server: Download file
    activate Server

    Process2->>File: Check if file exists
    File-->>Process2: File not found
    Process2->>Server: Download file
    activate Server

    Server-->>Process1: File downloaded
    Process1->>File: Save file locally
    deactivate Server

    Server-->>Process2: File downloaded
    Process2->>File: Save file locally
    deactivate Server
```
Notice that we download the same file twice!

The `fetch_file` function first checks if the file exists and has a hash that matches the known hash in the registry. If the file doesn't exists, or the hash does not match, then (and only then) it *acquires a file lock* to download or update the file. If the file exists and matches the expected value it simply returns the file path. The sequence diagram for the above looks like

```mermaid
sequenceDiagram
    participant Process1 as Process 1
    participant Process2 as Process 2
    participant Server as Registry
    participant File as Local File

    Process1->>File: Check if file exists and matches hash
    File-->>Process1: File not found or hash mismatch
    Process1->>File: Acquire lock
    File-->>Process1: Lock acquired
    activate File
    Process1->>File: Check if file exists and matches hash
    File-->>Process1: File not found or hash mismatch
    Process1->>Server: Download or update file
    activate Server

    Process2->>File: Check if file exists and matches hash
    File-->>Process2: File not found or hash mismatch
    Process2->>File: Acquire Lock
    Server-->>Process1: File downloaded
    deactivate Server
    Process1->>File: Save file
    File-->>Process1: File path on disk (returned by fetch_file)
    Process1->>File: Release Lock
    deactivate File
    File-->>Process2: Lock Acquired
    activate File
    Process2->>File: Check if file exists and matches hash
    File-->>Process2: File exists and matches hash
    Process2->>File: Release Lock
    deactivate File
    File-->>Process2: File path on disk (returned by fetch_file)
```
Notice that Process 2 checks the file again after it acquires the lock, and hence avoids the extra download introduced in the first sequence. The extra hash checks seem superfluous but they allow processes to skip lock acquisition if the file is already on disk, which should improve speed of file reads (which should be the majority of the calls to `fetch_file`.

lispandfound:
After a few fixes, I think this works! I have tested it in the `nshm2022-to-realisation` stage, and I use this code to define a realisation configuration block that looks like

```json
"input": {
        "commit": "dabeee2abe7497676c38ab1efed4882066c3a6df",
        "files": {},
        "registry": {
            "nshmdb.db": "1b73045b96390c30adc2a01b828a75adef588c5b456113f5607463854c19d91f"
        }
}
```

Here `commit` refers to a commit in the [ucgmsim/registry](https://github.com/ucgmsim/registry) repo (which is currently just a test repo while we wait to hear back about the suitability of eng-git). The `registry` key is populated from `registry.txt` in the GitHub repo. It can then download a slimmed down testing version of the NSHMDB and generate a rupture. I tested the locking feature by adding a breakpoint into the `registry.fetch_file` function to see that it blocked the other processes and released them once it had finished downloading the file. As expected, subsequent runs of `nshm2022-to-realisation` do not acquire locks because the hash of the database on disk matches the one in the registry.

qcore/registry.py
```diff
@@ -0,0 +1,117 @@
...
+    return qcore_pooch
+
+
+def fetch_file(
+    registry: pooch.Pooch, filepath: PurePath, lock_timeout: int = 60
+) -> Path:
+    """Return a path to the pooch registry file.
+
+    This function checks if pooch would download or update the requested
+    file path. If it does, then it acquire a lock on this file and pooch
```
AndrewRidden-Harper, line 93:
```suggestion
    file path. If it will, then it acquires a lock on this file and pooch
```

AndrewRidden-Harper (commented):
Small suggested change to a docstring text but otherwise looks great

AndrewRidden-Harper (dismissed):

qcore/registry.py
```diff
@@ -49,24 +60,58 @@ def qcore_registry(
     pooch.Pooch
         A Pooch object configured to manage the quakecore registry.
     """
-    if not reference:
-        reference = get_latest_registry_version()
+    if not registry:
+        if not reference:
+            reference = get_latest_registry_version()
```
joelridden, line 65: Is this meant to be "resolve_git_reference"?

joelridden (approved):

lispandfound:
@joelridden I think eventually this could also replace the downloading of the qcore resources file in places like `gmt.py`

https://github.com/ucgmsim/qcore/blob/be8aebd9f80406df6caab68aad96dbf6dfc3214d/qcore/gmt.py#L50-L66

This would mean changing to something like

```python
COAST_FILE = registry.fetch_file("qcore/data/Paths/coastline/NZ.gmt")
WATER_FILE = registry.fetch_file("qcore/data/Paths/water/NZ.gmt")
TOPO_FILE = registry.fetch_file("qcore/data/Topo/srtm_NZ.grd")
WATERNET_CHCH = registry.fetch_file("qcore/data/Paths/water_network/water.gmt")
# etc...
```

joelridden:
> @joelridden I think eventually this could also replace the downloading of the qcore resources file in places like `gmt.py`
> 
> https://github.com/ucgmsim/qcore/blob/be8aebd9f80406df6caab68aad96dbf6dfc3214d/qcore/gmt.py#L50-L66
> 
> This would mean changing to something like
> 
> ```python
> COAST_FILE = registry.fetch_file("qcore/data/Paths/coastline/NZ.gmt")
> WATER_FILE = registry.fetch_file("qcore/data/Paths/water/NZ.gmt")
> TOPO_FILE = registry.fetch_file("qcore/data/Topo/srtm_NZ.grd")
> WATERNET_CHCH = registry.fetch_file("qcore/data/Paths/water_network/water.gmt")
> # etc...
> ```
> 
> This would have the effect of versioning this data too, which I think would be a good idea.

I agree!

sungeunbae:
How do you plan to handle the lock if the file download process is disturbed/interrupted before fully completed, and it never gets a chance to release the lock?

> ```mermaid
> sequenceDiagram
>     participant Process1 as Process 1
>     participant Process2 as Process 2
>     participant Server as Registry
>     participant File as Local File
> 
>     Process1->>File: Check if file exists and matches hash
>     File-->>Process1: File not found or hash mismatch
>     Process1->>File: Acquire lock
>     File-->>Process1: Lock acquired
>     activate File
>     Process1->>File: Check if file exists and matches hash
>     File-->>Process1: File not found or hash mismatch
>     Process1->>Server: Download or update file
>     activate Server
> 
>     Process2->>File: Check if file exists and matches hash
>     File-->>Process2: File not found or hash mismatch
>     Process2->>File: Acquire Lock
>     Server-->>Process1: File downloaded
>     deactivate Server
>     Process1->>File: Save file
>     File-->>Process1: File path on disk (returned by fetch_file)
>     Process1->>File: Release Lock
>     deactivate File
>     File-->>Process2: Lock Acquired
>     activate File
>     Process2->>File: Check if file exists and matches hash
>     File-->>Process2: File exists and matches hash
>     Process2->>File: Release Lock
>     deactivate File
>     File-->>Process2: File path on disk (returned by fetch_file)
> ```
> Notice that Process 2 checks the file again after it acquires the lock, and hence avoids the extra download introduced in the first sequence. The extra hash checks seem superfluous but they allow processes to skip lock acquisition if the file is already on disk, which should improve speed of file reads (which should be the majority of the calls to `fetch_file`. I may end up simplifying this to always obtaining a file lock in the future.

lispandfound:
> How do you plan to handle the lock if the file download process is disturbed/interrupted before fully completed, and it never gets a chance to release the lock?

@sungeunbae the context manager ensures that, in the event an exception is thrown, the lock is always released with the python process. Under the hood, `filelock` uses the `flock()` syscall, which states in its [manual page](https://manpages.debian.org/bookworm/manpages-dev/flock.2.en.html) that the kernel will automatically cleanup these locks if the process dies unexpectedly. Of course we shouldn't rely on this, hence why filelock additionally uses the context manager. The combination of the two strategies should make deadlocks exceptionally rare. 

It's also worth noting that `flock` is *not* the same as creating a lock file like is done in other programs. It's an actual OS-level feature.

lispandfound:
@sungeunbae see this [SO answer](https://stackoverflow.com/a/12652718/4923537) that checks that perl (which uses the same `flock()` interface) releases locks in standard exit conditions, perl crashes and segfaults.

lispandfound:
Converted this PR to a draft while we want for the final Git host.

sungeunbae:
> @sungeunbae see this [SO answer](https://stackoverflow.com/a/12652718/4923537) that checks that perl (which uses the same `flock()` interface) releases locks in standard exit conditions, perl crashes and segfaults.

Excellent. You really thought this through.

AndrewRidden-Harper (approved):

PR #348, opened 2025-03-12
Use pooch for downloaded data

lispandfound:
This PR removes the download_data script and replaces it with [pooch](https://www.fatiando.org/pooch/latest/) for managing the data. Data is now downloaded on demand. That is you only download as much of the `qcore_resources.tar.xz` package as you actually use.

The change is entirely transparent to the user. If you want to use the `gmt.get_region` function, then the first time you run it will download the data to an os cache.

``` python
>>> from qcore import gmt
>>> # The above happens instantly because it no longer downloads data.
>>> gmt.get_region(172, -43)
Downloading file 'data/regions.ll' from 'https://www.dropbox.com/scl/fi/073atd0ebcrmob46a8yp5/regions.ll?rlkey=g54pfbd6jr25k24vm6ohgy6dq&st=1sgbox8p&dl=1' to '/home/jake/.cache/qcore'.
np.str_('NZ')
>>> # Subsequent calls will not download the file again.
>>> gmt.get_region(0, 0)
np.str_('CASCADIA')
```


joelridden (approved):

lispandfound:
The tests will fail because of the Jenkins user issue but I don't want to update the jenkinsfile with the other PR has the right changes already. GMT module is not tested anyway, so I'll override the check for this PR.

sungeunbae (approved):

PR #357, opened 2025-07-31
Vectorise and test broadband manipulation functions

lispandfound:
This PR:

1. Refactors the site amp modules to be easier to read and vectorises them with numba to work over multiple stations. The entry point `cb_amp_multi` now expects a dataframe of sites and returns an array with a row for each site and columns corresponding to amplification factors. This function no longer applies interpolation in the frequency space and expects the user to call the follow-up function `cb2014_to_fas_amplification_factors`. The primary reason for the latter change being to aid testing.
2. The bwfilter function is significantly cleaned up and appropriately documented. This documentation makes clear exactly how the frequency response at the cutoff frequency is supposed to behave. It was already vectorised, so there are no other changes here. I have removed the bandpass and passband filters because nobody was using them and I would need to construct tests for them (which takes some time).
3. The ampdeamp function has been improved significantly, with support for FFTW as the Fourier transform engine. This brings with it significant improvements to speed and multi-core support. It too now works for several sites at once.

I have added extensive testing to the `timeseries` and `siteamp` modules. The intention here is to bring them up to speed with the rest of the codebase and verify the changes are correctly implemented. Most importantly, the siteamp module now has an independent reference dataset of inputs and outputs, which the python version is tested to within 1% error. In the timeseries module, all of the promised guarantees about module behaviour (chiefly, signal attenuation in the frequency spectra, and the correct amplification of `ampdeamp`) are now tested explicitly.


lispandfound:
> Gemini is unable to generate a summary due to a potential policy violation.

Huh? Make it thinks the commit message

> [refactor: add an explicit CB2014 to FAS conversion function](https://github.com/ucgmsim/qcore/pull/357/commits/4dfa9b5d5acf0375562da23b57a92810b9229580)

implies something untoward about the code?

lispandfound:
Tests are failing because of the usual docker shenanigans. Shouldn't require any code changes to fix.

lispandfound:
Marked as draft as still a couple of extra changes to make before I am ready for review

qcore/siteamp_models.py
```diff
@@ -58,107 +52,553 @@ def amplification_uncertainty(
...
+    vpga_flat = vpga_arr.flatten()
+    pga_flat = pga_arr.flatten()
+
+    # Determine output size by running one case
+    # This is needed since we don't know the output size a priori
+
+    # Pre-allocate results array
+    results = np.zeros((n_cases, freqs.size), dtype=np.float64)
+
+    for i in range(n_cases):
```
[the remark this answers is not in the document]
lispandfound: This is not true and it is like copilot knows it too. The for loop is JIT-compiled and likely vectorised by the LLVM optimisations so that statement is almost certainly wrong. The `prange` comment is similarly wrong because `prange` is implicit for `parallel=True` decorated functions -- it should really only be used for functions where you need to control where the parallelisation occurs.

qcore/timeseries.py
```diff
@@ -5,308 +5,556 @@
...
+        the values of `amplification_factor`.
     """
-    nt = len(timeseries)
 
-    # length the fourier transform should be
-    ft_len = ampf.size + ampf.size
+    pyfftw.config.NUM_THREADS = cores
 
-    # taper 5% on the right using the hanning method
+    nt = waveform.shape[-1]
```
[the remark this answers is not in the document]
lispandfound: This is always true for valid files.

qcore/timeseries.py
```diff
@@ -5,308 +5,556 @@
...
-        ampf = 1.0 / ampf
-    # last value of fft is some identity value
-    fourier[:-1] *= ampf
+    # Amplification factor modified for de-amplification
+    if not amplify:
+        # Ensure ampf_modified is float32 for consistent operations.
+        # Handle potential division by zero if ampf contains zeros.
+        ampf_modified = np.where(
+            amplification_factor != 0, 1.0 / amplification_factor, np.inf
+        ).astype(waveform_dtype)
```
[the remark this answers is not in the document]
lispandfound: I consider this a negligible improvement.

qcore/siteamp_models.py
```diff
@@ -58,107 +52,553 @@ def amplification_uncertainty(
...
+)
+
+
+@njit(cache=True)
+def _cb_amp(
+    vref: float,
+    vsite: float,
+    vpga: float,
+    pga: float,
+    version: int = 2014,  # Changed to integer
```
joelridden, line 202: Probably don't need the comment here

qcore/siteamp_models.py
```diff
@@ -58,107 +52,553 @@ def amplification_uncertainty(
...
+    flowcap: float = 0.0,
+    freqs: np.ndarray = AMPLIFICATION_FREQUENCIES,
+) -> np.ndarray:
+    """
+    Numba translation of cb_amp.
```
joelridden, line 207: Do we have a function that explain what cb_amp is at all? Might be better for at least a short description with the addition of numba translation?

qcore/siteamp_models.py
```diff
@@ -58,107 +52,553 @@ def amplification_uncertainty(
...
+    n_cases = vref_arr.size
+
+    # Flatten arrays to handle both 1D and scalar inputs
+    vref_flat = vref_arr.flatten()
+    vsite_flat = vsite_arr.flatten()
+    vpga_flat = vpga_arr.flatten()
+    pga_flat = pga_arr.flatten()
+
+    # Determine output size by running one case
+    # This is needed since we don't know the output size a priori
```
joelridden, line 425: I think these can be removed now as the n_cases seems to come from the vref_arr.size now

qcore/siteamp_models.py
```diff
@@ -58,107 +52,553 @@ def amplification_uncertainty(
...
-    except IndexError:
-        pass
+    # Convert inputs to arrays and get dimensions
+    vref_arr = np.asarray(vref, dtype=np.float64)
+    vsite_arr = np.asarray(vsite, dtype=np.float64)
+    vpga_arr = np.asarray(vpga, dtype=np.float64)
+    pga_arr = np.asarray(pga, dtype=np.float64)
+
+    # Check that all arrays have the same shape
+    n_cases = vref_arr.size
```
joelridden, line 416: Maybe add the assert here

qcore/timeseries.py
```diff
@@ -336,34 +584,28 @@ def read_lfseis_directory(outbin: Path | str, start_sec: float = 0) -> xr.Datase
...
 
     # Set global attributes
-    ds.attrs["dt"] = dt
-    ds.attrs["resolution"] = resolution
-    ds.attrs["rotation"] = rotation
-    ds.attrs["units"] = "m/s"
+    ds.attrs["dt"] = header.dt
+    ds.attrs["resolution"] = header.resolution
+    ds.attrs["rotation"] = header.rotation
+    ds.attrs["units"] = "cm/s^2"
```
joelridden, line 608: Assuming the m/s to cm/s^2 change is intentional and correct

qcore/siteamp_models.py
lispandfound: yeah probably not

qcore/timeseries.py
lispandfound: Yeah m/s was a bug.

qcore/siteamp_models.py
```diff
@@ -58,107 +52,553 @@ def amplification_uncertainty(
...
-                        1.928, 2.194, 2.351, 2.460, 2.587, 2.544, 2.133,
-                        1.571, 0.406,-0.456,-0.82, -0.82, -0.82, -0.82, -0.82])
-    elif version == '2014':
-        # named c11 in cb2014
-        c10 = np.array([1.090, 1.094, 1.149, 1.290, 1.449, 1.535, 1.615,
-                        1.877, 2.069, 2.205, 2.306, 2.398, 2.355, 1.995,
-                        1.447, 0.330, -0.514, -0.848, -0.793, -0.748, -0.664,
-                        -0.576])
+
+@njit(cache=True)
```
sungeunbae, line 135: Is this meant to make it faster? Are we expecting to call this function repeatedly with the same input? Any risk associated with using cache?

lispandfound: So yes. The CB2014 model is called at every site three times during broadband simulation. The particular function `_compute_fs_value` is called twice for every frequency plus an extra two times. In Mike's case, this means this function is expected to be called `~20k (sites) * 22 (freqs) * 3 (components) = millions` of times. We used to get away with this with simulating every station simultaneously using multiprocessing in broadband simulation, which is why this was not vectorised before. But I would prefer to move away from multiprocessing for a variety of reasons, mostly for memory usage and speed. This implementation is considerably faster than the old multiprocessing version and has a more manageable memory profile because it doesn't spawn 64 copies of the python interpreter.

The top-level `_cb_amp` is expected to be called tens of thousands of times per simulation and can be run in parallel using `parallel=True` because sites are independent (this is what we were doing multiprocessing).

lispandfound: Actually, I may just remove the caching entirely.

qcore/timeseries.py
```diff
@@ -5,308 +5,556 @@
...
+            The shape of the resulting waveform array.
+
+        Returns
+        -------
+        np.memmap
+            A memory-mapped array containing floating point values
+            read from the seis file in the provided shape.
+        """
+        return np.memmap(
+            self.handle, dtype=self.f4, mode="r", offset=self.handle.tell(), shape=shape
```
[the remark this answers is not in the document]
lispandfound: It is a file-object with a `fileno()` so that's not a problem.

qcore/test/test_timeseries/test_timeseries.py
```diff
@@ -0,0 +1,862 @@
...
+    cutoff_freq_idx = np.argmin(np.abs(frequencies - cutoff_frequency))
+    high_freq_idx = np.argmin(np.abs(frequencies - high_frequency))
+
+    # Sanity checks, at each input frequency, we get the appropriate signal strength.
+    assert np.real(np.abs(fas[low_freq_idx])) == pytest.approx(
+        signal_length / 2, abs=0.001
+    )
+    assert np.real(np.abs(fas[cutoff_freq_idx])) == pytest.approx(
+        signal_length / 2, abs=0.001
+    )
```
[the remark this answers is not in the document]
lispandfound: Don't know what this comment is referring to.

joelridden (approved):

qcore/siteamp_models.py
```diff
@@ -58,329 +52,719 @@ def amplification_uncertainty(
...
+)
+
+
+@njit
+def _cb_amp(
+    vref: float,
+    vsite: float,
+    vpga: float,
+    pga: float,
+    version: int = 2014,
```
sungeunbae, line 202: nitpicking, but I'd think string is more flexible for version.

qcore/siteamp_models.py
```diff
@@ -58,329 +52,719 @@ def amplification_uncertainty(
...
+        The amplification factors `ampf0` interpolated to FFT frequencies
+        matching `dt` and `n`, and amplified according to the bandpass
+        filter `amp_bandpass`.
+    """
+    interpolated, ftfreq = interpolate_amplification_factors(freqs, ampf0, dt, n)
+    return amp_bandpass(interpolated, fhightop, fmax, fmidbot, fmin, ftfreq)
+
+
+@njit(
+    parallel=True,
```
sungeunbae, line 586: nice

